PR #209, opened 2014-08-28
make comment helper text optional

fishnix:
Add an option to display or hide the helper text in the comment section.


yellowled:
Has all this been merged? If so (or if we don't want to merge it for whatever reasons), the pull request should be closed. (Open pull requests make us look like bad maintainers.)

PR #210, opened 2014-08-30
Add s9ymarkup helper text option

fishnix:
Add an option to display (or not) the helper text in the comment section for s9y_markup


yellowled:
Has all this been merged? If so (or if we don't want to merge it for whatever reasons), the pull request should be closed. (Open pull requests make us look like bad maintainers.)

fishnix:
@yellowled I believe I solved my immediate problem with CSS as Garvin recommended.  While I  do think this could be useful, I'm OK with closing it and not merging (it looks like we've diverged quite a bit anyways!).  Cheers.

garvinhicking:
I'll close this, I believe we can't really merge this anyways. I'm still undecided whether such an option would be good or if the CSS-approach is "good" enough. So I hope that's ok :)

Issue #279, opened 2015-01-26
multiple plugin updates - do not fallback


ophian:
Stick onto update page after plugin update submit, as long as there is more to do (say, more updates are available).

yellowled:
You mean instead of going back to the *Plugins* page, go back to *Update [Group] plugins* page if there are more updates available?

+1

bauigel:
>
> You mean instead of going back to the *Plugins* page, go back to *Update
> [Group] plugins* page if there are more updates available?
>
> +1
>

​+1​

onli:
I rather think we should have a button there  to update all plugins at once. That can be done with javascript that clicks the buttons and waits till the update is doen before clicking the next, or in core or spartacus with a function queuing the updates. Both ways are valid and should be easy enough, this modification here would probably even be more difficult.

Also, to sometimes go to the plugin page or sometimes not does not sound good/consistent, even though I get why this might be helpful in this specific case.

yellowled:
> I rather think we should have a button there to update all plugins at once.

What if you don't *want* to update **all** plugins at once? Then it would still be nice to be redirected back to the update page in case you'd want to update multiple plugins, but not all of them.

onli:
Did you ever want to not update all of them? :)

We could make them selectable, but I really doubt that case happens often enough to warrant the effort. Given that the "update all" approach works. If not, to try to be smart to which page we go afterwards would be better than now I guess.

yellowled:
> Did you ever want to not update all of them?

Yes.

For example, we have been holding off on updating the EasyPodcasting plugin in the InfoCamp blog. I forgot why, but there was a reason.

garvinhicking:
IMO this should be something for 2.x.0, not a point 2.0.1 release. 2.0.1 should only get well-tested bugfixes, but everything that could create probable trouble should be put to 2.1.0 (were we already have somet hings like that)

yellowled:
Agreed, but I don't think “well-tested” is a good criterion for this. This *is* well-tested, but it's not a bugfix.

onli:
Let's follow http://semver.org/
2.0.x: bugfixes
2.x.0: new features
X.0.0: Updates with bc-breakage/API-change

yellowled:
For issue assignment, the semver policy is fine, but I think it's too complicated for us in terms of branch management. So we should stay with `master`, `2.0` and feature branches for bigger stuff that might break things or be thrown away after discussion.

[...]

PR #289, opened 2015-02-04
Feature update all

onli:
This shall solve #279 by adding an update all button to the update page.


garvinhicking:
@onli I think we should merge this branch with master (2.1.x)? Not sure how this can be done though?!

onli:
Not an expert either, but I think one basically only has to be in master and then make a `git merge` with the feature branch, and then edit away the conflict markers. You saw the "Use the command line" help link?

If you prefer I could try to do the merge myself.

garvinhicking:
@onli That would be great, as I'm a bit short on time

yellowled:
> edit away the conflict markers

Make sure to not only edit out the markers but also check the code differences *around* those markers. ;)

onli:
That only kind of worked. The commit got really big and does show more than the diffs. Sorry, don't know what went wrong. 

Is the endresult working for you?

ophian:
You should revert that and first learn your git installation or your code editor the correct charset encoding <sup>(1)</sup> and/or line ends <sup>(2)</sup> , I assume. Check this out.
I think this is what YL has discovered several times in the last few weeks, when he commited that some files were touched in whole, since they had a wrong encoding or line end, most probably while they were touched by you before.

<sup>1)</sup> It should be UTF-8 without BOM for all lang/utf8/`*.*` files, but also for the NEWS file and some more. The editor should do that automatically.
<sup>2)</sup> https://help.github.com/articles/dealing-with-line-endings/

yellowled:
I was always under the impression that *all* files which are *not* in a UTF-8 directory should be iso-8859-1 (including, for instance, `docs/NEWS`). If that is not true, we need a file or wiki page listing all files that need to be UTF-8. No contributor can be expected to remember those by heart.

ophian:
Most others are ANSI yes, but some are set to UTF-8w/oBOM, see NEWS for example. (Could be easily converted though.)
Oh and I  had a deeper look. Onli has now commited a whole bunch  of new `*.orig` files (most to live in UTF.8/ lang dirs as ANSI), eg NEWS.orig or plugin_lang.php.orig, etc. See
https://github.com/s9y/Serendipity/tree/7c74930809f1e86b434a5e2814c519c1a2cf2c6b/lang/UTF-8

yellowled:
Yeah, I'm not sure they're supposed to be UTF-8, though.

As for `*.orig`, yes, they need to be removed from the remote repository. It's pretty easy to still keep them in a local repository, though. Just add `*.orig` to `.gitignore`.

onli:
> Just add *.orig to .gitignore.

Good idea. I used meld to do the merge, as there were quite a lot conflicts (every fucking language file), didn't catch that.

The other stuff with the encoding however sucks. I'm not using something special in my setup, and I won't be able to make this work. @garvinhicking, the whole encoding stuff seems to work for you, would you be so kind and do the merge?

And after that we should move everything to UTF-8 and unix-newlines and solve those problems one and for all.

yellowled:
Depending on what editor you use, using an [.editorconfig](http://editorconfig.org) might help.

I literally do everything else in utf-8, so I have it set to latin1 for s9y repositories. Just have to remember to save the ones that are supposed to be utf-8 in utf-8.

That being said, I'm all for moving on to utf-8, but I think we discussed that already and there was a reason why that was not possible?

onli:
> Depending on what editor you use, using an .editorconfig might help.

Like I said, as far as I there were files in the commit I  did not open in any editor ever.

> ut I think we discussed that already and there was a reason why that was not possible?

Yes and no. We discussed it in the context of language files, where we have to re-encode them into their native charset when we restore the old option, wehich needs to be tested and I never implement. There is no reason at all to not have UTF-8 in things like doc/NEWS.

garvinhicking:
The thing is this: We only use ISO-8859-1 characters in 99% of all 
files, so it shouldn't matter if editors save a file as UTF-8 or ISO-8859-1.

What DOES make a difference is line termination / EOL. Ideally, an 
editor should never autocorrect anything there, at least this is how I 
configure my editors. Everything actually should already be 
unix-newline, I use nothing else.

This whole merging is a bit complicated to me (the 2.0 -> master merge I 
did was not a merge, but a complete renaming of a branch, this would 
only work if we merged everything back from the feature branch with 
master and then rename the feature branch to master and now my head 
hurts already). I think we shouldn't do this kind of feature branching 
in the future but better work with local commits that are only pushed to 
master when they are ready to be shown - and until then, only 
post/provide diffs for people to check out the feature? Don't know. It 
really sucks. :(

garvinhicking:
Since master and feature_updateall branche diverge quite a bit already, 
I honestly wouldn't know how to perform a merge.. I probably would need 
to checkout both branches, make a diff, and manually sort through all 
things and only keep those that were changed for feature_updateall. But 
since I don't know that new code, I would very well miss a lot of things.

Bottom line, I don't think I have the bells&whistles to perform this at 
this point.

yellowled:
> it shouldn't matter if editors save a file as UTF-8 or ISO-8859-1

So, rule of thumb: it does not matter whether a file utf-8 or iso-8859-1, as long as it does not contain characters that *require* a specific encoding (like the German lang files because of German umlauts, for example)?

> Ideally, an editor should never autocorrect anything there

Well, if the editor is configured to use Unix newlines (LF), that should be okay (that's what I use).

> I think we shouldn't do this kind of feature branching in the future but better work with local commits that are only pushed to master when they are ready to be shown - and until then, only post/provide diffs for people to check out the feature?

I think it should not really make a difference if the feature branch is local or remote. If it diverges from `master`, the merge will have potential conflicts. Keeping the feature branch local just makes it more difficult for others to test.

onli:
> Since master and feature_updateall branche diverge quite a bit already,

They did not :)

The merge is pretty easy to do. You would just pull both branches, go into master and do `git merge --no-ff feature_updateAll`. Then there are conflicts, but all the conflicts are easy. In the news-file, we want to mix the changes together, in css and js it is only adding the stuff from the feature-branch, and in the lang-fiels it is again mixing the changes together. Thing is, you would call `git mergetool -t meld` or whetever tool you want to use to do the merge and it would highlight every conflict for you and lead you through that. With that, it's a matter of 15 minutes max.

Only afterwards there was the problem, as I did a `git add .` and got this disaster commit that was not visible for me in the commit log as well, as it was only listing the files there with the note that it was merging a conflict. 

That is why I see no way for me to fix this. But the merge itself is easy.

onli:
Give me some minutes, I have an idea and will try it myself.

yellowled:
> in css and js it is only adding the stuff from the feature-branch

Nope.

```
<<<<<<< HEAD
.hidden {
=======
.hidden,
.msg_error > br,
.msg_success > br,
.msg_notice > br,
.msg_hint > br,
.no-cssanimations .animated-css {
>>>>>>> feature_updateAll
```

This should be

```
.hidden,
.no-cssanimations .animated-css {
```

after the merge.

onli:
Don't worry, I saw that in the diff and only forgot to list it.

yellowled:
Just trying to help. You're on your own with the lang files, though. That shit scares me now. ;-)

onli:
7740886ff9ebf74d0365c9d69eb9b46c2876b077 looks better :) Could you all please have a look?

> Just trying to help.

Thanks :) It is annoying when stuff like that does not work immediately. Only merging some lines… Even the lang files are only saying "keep the 4 lines from master and add the 2 from the feature branch below".

@yellowled, could you have an extra look at the admin/style.css? The merge that was not a conflict looks like it changed a bit more than intended, and I'd prefer not to revert now…

yellowled:
> The merge that was not a conflict looks like it changed a bit more than intended

At first glance, that looks as if it deleted most of the stuff I worked on over the past few weeks.

onli:
Have a second ;) If need be I'd restore that manually (by taking the last from master and copying the updatebutton changes in manually).

yellowled:
Concerning `/templates/2k11/admin/styles.css`, it's actually pretty straight-forward:

* All that was removed (is red) in 7740886ff9ebf74d0365c9d69eb9b46c2876b077 needs to be in the file
* All that was added (is green) in 7740886ff9ebf74d0365c9d69eb9b46c2876b077 needs to be removed **except for** lines 322-353 (`progress` polyfill) and lines 2434-2478 (styles for the `progress`/update stuff)

Can't say much about the rest. `serendipity_editor.js.tpl` and `modernizr-2.8.3.min.js` look okay.

yellowled:
Side note: Did you not do a `git pull` on `master` before trying to merge? This looks way more off than it should be.

onli:
> Did you not do a git pull on master before trying to merge? This looks way more off than it should be.

I actually removed my local repo before and got a new one from github. No mistake there.

All other files look ok as well, I checked. It's only the styles.css that is mixed up.

> it's actually pretty straight-forward:

Good. Are you already doing that or shall I?

yellowled:
> Are you already doing that or shall I?

I'm holding on to my local repo with the correcty stylesheets, please go ahead. :)

onli:
Good idea to hold on to them…

I tried it with https://github.com/s9y/Serendipity/commit/e7502ef9bcf2ef29e25e23e7b59ed1e5df4de89e. Hard to control for me, but seemed to work?

yellowled:
Yup, perfect. Well, I think the diff showed a few unnecessary spaces, but my next commit for the file is going to remove those automagically anyway, so I would not bother. (`.editorconfig`, `trim_trailing_whitespace`)

onli:
Ok, thanks everyone, and sorry for the hickup.

PR #367, opened 2015-09-22
Feature index cleanup

onli:
This is a refactoring of the index.php. It moves the routing code into the new file **include/functions_routing.inc.php**. The index.php is then very much smaller and the flow to genpage easier to understand.

Note that Garvin objected the removal of the calender and the comments by author overview. I disagree with that – imho, they should be moved into plugins if they are really used. But that can't happen under the current path before a external routing framework (like http://flightphp.com/) is used that can be modified by those plugins, which sure won't happen as long as this cleanup is not in master. Which means I reverted those removals.

Might still introduce some breakages.

Discussion thread: http://board.s9y.org/viewtopic.php?f=11&t=20467


ophian:
To make up my mind on this. Please let me ask, since I did not really get this so far:
Is there a real benefit for this cleanup approach, beside that YOU personally think it is better to understand or better readable?
Is it faster, better usable, preparing for tomorrow, or something else? Where does this head to?

And since we are on this; I found this some weeks ago and always wanted to fix it:
https://github.com/s9y/Serendipity/blob/4f2ea4de8a70825b15e357fb3bbf5a76cb178c5d/index.php#L17
The doku says that `$HTTP_SERVER_VARS` were set deprecated with PHP 5.3 and removed with 5.4!
Now, does this line need to just use $_SERVER instead? And which looped header is actually the one being used in modern systems? @garvinhicking

onli:
> Is it faster, better usable, preparing for tomorrow, or something else? 

Please check it out and look at the code. I doubt you will disagree that it is better readable. It also removed some old code, so it might be faster despite the added function calls.

> Where does this head to?

I described that in the forum thread. In the end, I would like to use a web framework like http://flightphp.com/ instead of letting s9y do everything manually. This cleanup however does have a value on its own even if we don't go there.

> The doku says that $HTTP_SERVER_VARS were set deprecated with PHP 5.3 and removed with 5.4!

I would not mind if you commit a fix, and I see why you connect it with the index cleanup. But still, if we need to discuss it further than Garvin saying *yeah, change it to $SERVER*, please do it in a separate issue.

garvinhicking:
yeah, change it to $_SERVER :)

onli:
@garvinhicking Can this be merged?

onli:
I merged that now. Review and revert if needed, but it should work fine.

garvinhicking:
@onli I don't seem to be able to find a way to properly review this; do you maybe have the time to sum up, what in the final instance you have removed (be it because it was broken or you think it should better be a plugin).

And could you also document in docs/NEWS what you changed, and if it has implications for plugin / theme authors or any BC breaks, to name them? That would also make it easier for me to get up to speed on this.

onli:
That is the thing, I'm not aware of any such changes. It is all under the hood and should have no implications for no one. That is because I reverted the removals you objected to. The only thing left is the multi-author option, https://github.com/s9y/Serendipity/commit/55ae435c898bfcff1a4722a34e25116188a1079c. But that I did not break, it was broken before, I just removed the dead code. Shall I document that?

garvinhicking:
Yeah, it would be good to mention something like "Rewrote Routing code for index.php to be outsourced into include/functions_routing" and "Removed broken feature for viewing blog entries by multiple authors".

onli:
That describes it well :) I commited it

PR #372, opened 2015-11-01
additional space in TopExits/Referrers for better word-wrap

erAck:
There was no space between total value and next host, so display,
depending on available width, was

www.example1.com (12)www.example2.com
(23)www.example3.com (34)

With an additional trailing space after total it is

www.example1.com (12)
www.example2.com (23)
www.example3.com (34)


yellowled:
@erAck Good catch, thanks!

yellowled:
**Note:** I also `cherry-pick`ed this to `2.0`.

PR #390, opened 2016-02-08
Disable email notifications for replies send from backend

xoxys:
Hi,
this problem was discussed with onli on s9y forum. Every time if an author or admin reply to an comment from the backend, he receives an notification email. I would like to disable this behavior because the author knows that he has send a reply and don't require a notification for this topic.

Thanks a lot


onli:
And merged. Thanks a lot for contributing your changes :)

PR #391, opened 2016-02-09
check session variable to detect a reply from author

xoxys:
Hi,
i removed the variable $backend and use $serendipity instead to detect a replay from the author. So the verification works also in the frontend if a user is logged in.


onli:
Looks okay to me. I'll merge it. Thanks again!


Just because it got me thinking: 

    if ($serendipity['serendipityEmail'] != $row['email']) {

Would that work already?

xoxys:
I have tested the code and for me it works. But i am not completely sure if this is the best way. There are two fields in the array $serendipity['serendipityEmail'] and $serendipity['email'] and both have the same value... But i will look deeper to it tomorrow.

PR #392, opened 2016-02-12
Provide authenticated userdata for frotend comment form 

xoxys:
Hi,
this patch will provide the email and realname of an authenticated user and commmit these values to commentform function. Now templates can pre-fill comment form with data from logged in users.


onli:
Thanks!

ophian:
What about those which do no want pre-filling? Is it configurable? For the latter, what is the default then?

onli:
Ian, this is linked to a forum discussion, http://board.s9y.org/viewtopic.php?f=10&t=20535&start=30. Please discuss there.

Also, please look at the code. What is the default is not a sound question here.

PR #406, opened 2016-05-09
Show two "comments feed" buttons when both feed formats (RSS+Atom) activated

blackwarthog:
Show two "comments feed" buttons when both feed formats (RSS+Atom) activated and "subToMe" button disabled.

Sidebar before:
    RSS Feed
    Atom Feed
    Comments

Sidebar now:
    RSS Feed
    Atom Feed
    Comments (RSS)
    Comments (Atom)

When only one format activated "comments" button shows without format name, for ex:
    RSS Feed
    Comments

### DIFF
--- a/plugins/serendipity_plugin_syndication/serendipity_plugin_syndication.php
+++ b/plugins/serendipity_plugin_syndication/serendipity_plugin_syndication.php
@@ -203,10 +203,18 @@ class serendipity_plugin_syndication extends serendipity_plugin {
         }
 
         if (serendipity_db_bool($this->get_config('show_2.0c', false)) || serendipity_db_bool($this->get_config('show_comment_feed', false))) {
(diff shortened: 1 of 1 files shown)


onli:
I'm not sure that's a good idea. Atom is dead, there is no need for it anymore. We already moved the defaults of the syndication plugin for the entry feed to RSS only. I see no gain in making it possible to show the atom comment feed, I'd rather remove those altogether (which we can't because people still might be subscribed to them).

On the other hand, making it possible is not the same as making it default. If the current behaviour were unexpected we maybe should accept this PR. The code looks good.

I'm undecided. Maybe someone else wants to chime in?

plugins/serendipity_plugin_syndication/serendipity_plugin_syndication.php
```diff
@@ -1,9 +1,5 @@
 <?php
 
-if (IN_serendipity !== true) {
```
onli: Small remark: I don't think that should be removed here, as it is present in the other plugins.

blackwarthog: Rly, i should to turn it back. I didn't want to change these lines, it's mistake.

PR #407, opened 2016-05-09
Show two "comments feed" buttons when both feed formats (RSS+Atom) activated

blackwarthog:
Show two "comments feed" buttons when both feed formats (RSS+Atom) activated and "subToMe" button disabled.

Sidebar before:
RSS Feed
Atom Feed
Comments

Sidebar now:
RSS Feed
Atom Feed
Comments (RSS)
Comments (Atom)

When only one format activated "comments" button shows without format name, for ex:
RSS Feed
Comments

----

Apologies for troubles in previous request

I really don't know how atom popular or not. My aggregator supports both RSS and Atom. This patch just makes plugin more whole and completes some functionality which wasn't fully implemented.


onli:
No problem. Again, thanks for contributing to serendipity. 

I think you are right, it is more complete this way. I'll merge it.

Issue #431, opened 2016-12-11
HTTP 500: /rss.php?version=rubbish


gnuheidix:
I've noticed in my php error log that this path gets hammered by SQL-Injection bots and they achieved to make s9y crash continuously.
I've noticed that as soon as rss.php gets called with anything that doesn't fit the template file loading mechanism, it crashes. I've tried some attacks but I was unable to retrieve other content than RSS templates from the filesystem.

example path: /rss.php?version=justatest
s9y version: 2.1-beta3
`PHP Fatal error:  Uncaught  --> Smarty: Unable to load template file 'feed_justatest.tpl' <-- 
  thrown in /path/to/s9y/bundled-libs/Smarty/libs/sysplugins/smarty_internal_template.php on line 219`

I suggest to check the version parameter very thoroughly and only allow a certain set of characters being used for file system access. e.g. only [a-z0-9.]
Furthermore, in case the template file doesn't exists a HTTP 404 gets returned.

onli:
Thanks for the report, good catch. For the log, https://github.com/s9y/Serendipity/blob/3179ed7d579209ffe25f71a37b24d0e70536cc3c/rss.php#L242 is how that parameter is used currently. I propose to limit the allowed final filenames to existing feed template files.

gnuheidix:
Thanks for the clarification, so the parameter is already being sanitized properly.

Checking the existence of the template file and rendering a 404 should be enough to fix this bug. I found a Smarty method [1] which could help us probably.

[1] http://www.smarty.net/docsv2/de/api.template.exists.tpl

onli:
termplate_exists is the right idea here (rss.php is actually one of the few places where our custom fallback chain is not used, if I see that right), but I prefer to close that up even more. Such access to the filesystem is critical.

e9f98f8 explicitly allows only access  to the templates known to exists, and is in line how s9y behaves when detecting the file version automatically from the path, in `serendipity_discover_rss`. It would be great if you could further test that commit (it seemed to work fine on my dev system).

gnuheidix:
I had to make the code a little bit stricter. see #432 

The code maintainability is still not great because all known feed types are now in rss.php and in `serendipity_discover_rss`. ...but on the upside, it doesn't crash anymore :-)

onli:
Yeah, it could have been nicer to refactor that. But on the other hand the supported feed formats did not change in around 10 years, so that had no priority for me.

[...]

PR #432, opened 2016-12-14
Strict feed template name checking (#431)

gnuheidix:
otherwise requests like /rss.php?version=1 will still crash


onli:
Ah. Because 1.0 equals 1. Thanks.

PR #446, opened 2017-02-07
makes sure that only strings are being processed in searchTerm

### LINKED ISSUE #444: HTTP500: search term crashes on array
The search request can be done in a way, so that s9y crashes.

Request: GET /index.php?serendipity[action]=search&serendipity[searchTerm][]
Version: 2.1-rc1
Errorlog:
```
PHP Fatal error:  Uncaught ErrorException: Warning: strip_tags() expects parameter 1 to be string, array given in serendipity_config.inc.php:430
```

gnuheidix:
strip_tags crashes in case it's getting an array

fixes #444


onli:
Works for me, thanks for fixing!

Issue #445, opened 2017-02-06
HTTP500: comment preview and submit crash on array


gnuheidix:
The POST request which adds a comment can crash s9y in case the comment field gets modified into an array.
Request example: POST /index.php?url=archives/63-blablabla.html

Form-Parameter example 1: - see _serendipity[comment][]_ - preview used
![form_params](https://cloud.githubusercontent.com/assets/234757/22659968/f26f06dc-ec9f-11e6-85a5-99164dc24b33.png)
Form-Parameter example 2: - see _serendipity[comment][]_ - submit used
![form_params2](https://cloud.githubusercontent.com/assets/234757/22660477/f89a2f1c-eca1-11e6-8dd7-5144716d4f63.png)

Version: 2.1-rc1
Errorlog for example 1:
`PHP Fatal error:  Uncaught ErrorException: Warning: htmlspecialchars() expects parameter 1 to be string, array given in /srv/gnuheidix.de/include/compat.inc.php:516`

Errorlog for example 2:
`PHP Fatal error:  Uncaught ErrorException: Warning: trim() expects parameter 1 to be string, array given in /srv/gnuheidix.de/include/functions_routing.inc.php:342`

onli:
I also saw #444.

In both cases my questions is: Can this happen in the normal UI? I think I miss the problem with those two crashes. Of course, we should handle all situations/all variations of input variables that might arise in normal use, and if normal use can cause this, then we need to catch it. But when PHP crashes because some input variable got modified, to my understanding that is not a problem at all, and also not something that can be prevented with reasonable effort.

Can you elaborate a bit why those two crashes are problematic? And @garvinhicking, please shut me up if I'm  spreading nonsense, not exactly sure I'm not missing something here.

gnuheidix:
A well built web application should never respond with a 500 if somebody fiddles with user input. This should be handled by the input validation.
A 500 is only ok in case the server is low or out of on resources or a required external system like a database or so is unavailable.

Furthermore, in PHP web applications Fatals tend to produce stacktraces visible to users. That leaks security relevant system information depending on the level of (mis)configuration in php.ini.

onli:
We actually try to control the amount of error output. In production mode, no stacktrace should be visible, if the user does not explicitly deactivate our configuration.

[...]

PR #448, opened 2017-02-08
HTTP500: crashing when username is an array

gnuheidix:
Similar to #444 and #445, the login functionality crashes when the handed in username is an array.

Error:
`PHP Fatal error:  Uncaught ErrorException: Warning: PDO::quote() expects parameter 1 to be string, array given in include/db/pdo-sqlite.inc.php:122`

Request: POST /serendipity_admin.php
Version: 2.1-rc1
Form-Input needed to break it: _serendipity[user][]_

### DIFF
--- a/include/functions_config.inc.php
+++ b/include/functions_config.inc.php
@@ -530,7 +530,7 @@ function serendipity_authenticate_author($username = '', $password = '', $is_has
     }
 
     if ($debug) fwrite($fp, date('Y-m-d H:i') . ' - Login username check:' . $username . "\n");
-    if ($username != '') {
+    if ($username != '' && is_string($username)) {
         if ($use_external) {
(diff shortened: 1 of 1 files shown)


onli:
Thanks for putting this into a PR! I would like to wait with merging this till after 2.1 is released. The change is small in code and looks sound, but the login function is very important and we just might miss some voodoo interaction here. It should go through a beta or at least an RC, and I'm not sure we'll have another proper RC for 2.1. If we have, I'll try to hit the merge button before.

onli:
Okay, the time to merge this is now. With this, the login will cleanly fail if the form is manipulated. And the code behind the if clearly expects a string. However, what happens if the username is 12345, which can later on be safely treated as string but is none? In my tests such a username was already a string when being given to `serendipity_authenticate_author`, so that's fine as well. But we'll have to keep this in mind when rewriting the autologin.

PR #461, opened 2017-03-14
Fixed uninitialized Smarty value (fixes #460)

### LINKED ISSUE #460: Upgrader reports a wrong directory path
When performing an upgrade the upgrader checks some directory permissions. One of them is the $uploadHTTPPath. The check is working correctly but if the check is missing write permissions the displayed path is wrong. In my case it reported the path to be "/var/www/htdocs/" instead of  "/var/www/htdocs/uploads/".
The reason for this is, that the Smarty variable $uploadHTTPPath isn't assigned. The responsible Smarty code is the following line:
```smarty
# templates/2k11/admin/upgrader.inc.tpl, line 112
<dt>{$basedir}{$uploadHTTPPath}</dt>
```
So if the Smarty variable isn't assigned the displayed path only consists of $basedir. To solve this simply assign the variable to the correct value.

webcompas:


yellowled:
(We can of course merge if *one* of you approves after review.)

yellowled:
@webcompas Thanks! 👍

PR #468, opened 2017-04-08
Dashboard: show as many drafts as configured.

### LINKED ISSUE #465: Dashboard shows 5 comments, but 4 drafts
To reproduce:

1. Have at least 5 recent comments
2. Have at least 5 drafts

Expected behavour: dashboard shows 5 recent comments and 5 drafts.
Actual behaviour: dashboard shows 5 recent comments, but only 4 drafts.

th-h:
The dashboard will show up to $dashboardLimit future
entries; if the number of future entries is still
less than $dashboardDraftLimit, it will add drafts
up to $dashboardDraftLimit.

If there are _no_ future entries, $entries is no
array, but has a value of "1"; count($entries)
will then be one, too, so one draft less will
be shown.

Closes #465.

### DIFF
--- a/include/admin/overview.inc.php
+++ b/include/admin/overview.inc.php
@@ -104,7 +104,10 @@ $entries = serendipity_fetchEntries(
                      'e.timestamp >= ' . serendipity_serverOffsetHour() . $efilter
                    );
 
-$entriesAmount = count($entries);
+$entriesAmount = 0;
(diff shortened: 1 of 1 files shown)


onli:
Taking this on since we went through how the count-code should behave at the mdinight debugging. PR seems sound to me and works.

Issue #469, opened 2017-04-08
Add all variables to serendipity_config_local.inc.php


th-h:
There are many variables one can set in `serendipity_config_local.inc.php`, see https://docs.s9y.org/docs/developers/code-primer.html#docs-initializing-the-framework-serendipity_configincphp-and-serendipity_config_localincphp

It would be nice to have all that variables documented in `serendipity_config_local.inc.php`, with their default values.

garvinhicking:
Those variables are actually all set in serendipity_config.inc.php - if we add them to the local config as well, we would have to redundantly add them at two places and possibly change defaults at two places, which I don't really like. We can't ONLY put them to the local file, because it's not maintained in git. Soooooo....We might want to add a link to the documentation page in the config_local file instead? :-)

th-h:
I don't think so - some variables aren't set in `serendipity_config.inc.php`, e.g. `$dashboard*Limit`. I'd like to add those, commented out, so you can see everything you can change.

The defaults that are set in `serendipity_config.inc.php` can and should stay there, I think.

th-h:
After thinking some more, let's add a link to the documentation page to `serendipity_config_local.inc.php` and move all defaults to `serendipity_config.inc.php` that are not yet there (see PR 493 for example).

PR #480, opened 2017-04-08
Make backend page title order configurable.

### LINKED ISSUE #413: Change Title of Backend Page
Please change the title of the backend page (shown with BLOGURL/serendipity_admin.php) to something that reflects the blog title.

Should be different from "Startseite | Serendipity Verwaltungsoberfläche", maybe "BLOGTITLE | Admin"?

th-h:
The default page title of backend pages is
"section | blog title | $admin". You can
now change the order to "blog title | section
 | $admin" by setting $serendipity['title_first']
to "blogtitle" in serendipity_config_local.inc.php

Closes #413

### DIFF
--- a/serendipity_admin.php
+++ b/serendipity_admin.php
@@ -227,9 +227,10 @@ if ($ajax) {
         $admin_vars[$poll_admin_var] =& $$poll_admin_var;
     }
 
-    $admin_vars['out']       = array();
-    $admin_vars['no_create'] = $serendipity['no_create'];
-    $admin_vars['title']     = $admin_section;
+    $admin_vars['out']         = array();
+    $admin_vars['no_create']   = $serendipity['no_create'];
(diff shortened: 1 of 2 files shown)


onli:
Since https://github.com/s9y/Serendipity/issues/469 is open I will merge without addressing it here, but in general, serendipity_config.inc.php should contain the new variable (whose name does not make clear it is for the backend, naming is hard, but I think we can improve that) with a sound default value.

th-h:
Yes, I'm not very fond of the name myself, had been trying to keep it short ...

What about "backend_title_first" or "backend_title_order"?

onli:
backend_title_first if it is a boolean (and that would be okay), default true. backend_title_order if it contains multiple possible orders that are checked. I think the first variant is nicer, we won't have multiple orders, and can still change if that turns out to be wrong.

PR #489, opened 2017-04-12
Change Spartacus default mirror to github.

th-h:
Netmirror is not very reliable currently.

Add Changelog file.

### DIFF
--- a/plugins/serendipity_event_spartacus/ChangeLog
+++ b/plugins/serendipity_event_spartacus/ChangeLog
@@ -1,0 +1,4 @@
+2.37.3:
+-------
+    * Change mirror order (new default: github)
+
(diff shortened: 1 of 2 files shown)


onli:
Necessary. Possible that we'll have to remove netmirror completely and add an upgrade-task.

garvinhicking:
I am a bit afraid to use github by default, because I don't really fully know if what we're doing there is something they really allow us to do, so it could change any time.

netmirror is of course something that is currently not working, I tried to reach my contact there, but couldn't get a response yet.

Maybe we should default to s9y.org as our mirror...?

th-h:
> Maybe we should default to s9y.org as our mirror...?

It was last in the list before, so I thought perhaps you wanted to save on traffic to the site. Otherwise, it would be the logical choice.

onli:
I don't think that using github is problematic. There were articles about the performance impact of a whole programming language package manager (looked it up, it was cocoapod, see http://blog.cocoapods.org/Master-Spec-Repo-Rate-Limiting-Post-Mortem/). Our traffic will be minimal compared to that :)

garvinhicking:
Yeah the thing about s9y.org always was that I had the feeling it broke more often than netmirror.... :-D

Well...we'll see I guess...

PR #495, opened 2017-04-13
Add Net/DNS.php 1.0.7 to core.

### LINKED ISSUE #459: Bundled Net/DNSBL.php depends on missing Net/DNS.php
I am using Serendipity 2.1-rc1 with PHP 7.0.15-0ubuntu0.16.04.4.

In my webserver’s error log, I am finding error messages like the following (paths and URLs anonymized):
```
[Wed Mar 08 12:37:03.609101 2017] [proxy_fcgi:error] [pid 27470:tid 140017611503360] [client 46.118.153.31:51768] AH01071: Got error 'PHP message: PHP Warning:  Uncaught ErrorException: Warning: require_once(Net/DNS.php): failed to open stream: No such file or directory in /s9yroot/bundled-libs/Net/DNSBL.php:34\nStack trace:\n#0 /s9yroot/bundled-libs/Net/DNSBL.php(34): errorToExceptionHandler(2, 'require_once(Ne...', ‚/s9yroot...‘, 34, Array)\n#1 /s9yroot/bundled-libs/Net/DNSBL.php(34): require_once()\n#2 /s9yroot/plugins/serendipity_event_spamblock_rbl/serendipity_event_spamblock_rbl.php(89): require_once(‚/s9yroot...')\n#3 /s9yroot/include/plugin_api.inc.php(1139): serendipity_event_spamblock_rbl->event_hook('frontend_saveCo...', Object(serendipity_property_bag), Array, Array)\n#4 /s9yroot/include/functions_comments.inc.php(1006): serendipity_plugin_api::hook_event('frontend_saveCo...', Array, Array)\n#5 /s9yroot/include/functio...\nPHP message: PHP Fatal error:  serendipity_event_spamblock_rbl::main(): Failed opening required 'Net/DNS.php' (include_path='.:/usr/share/php:/s9yroot/bundled-libs/:/s9yroot/bundled-libs/Smarty/libs/:/s9yroot/:') in /s9yroot/bundled-libs/Net/DNSBL.php on line 34\n', referer: <anonymized>
```

I believe this happens because the bundled `Net/DNSBL.php` depends on `NET/DNS.php` which for some reason is not bundled with s9y.

I cannot tell for sure whether this error first appeared after upgrading to s9y 2.1-rc1, upgrading to PHP 7, or has already been present for a while. Anyway, the underlying problem still seems to be present in the current version from Git.

th-h:
Fixes #459

See https://pear.php.net/package/Net_DNS/

### DIFF
--- a/bundled-libs/Net/DNS.php
+++ b/bundled-libs/Net/DNS.php
@@ -1,0 +1,509 @@
+<?php
+/**
+* Class to provide IPv4 calculations
+*
+* Module written/ported by Eric Kilfoil <eric@ypass.net>
+*
+* This is the copyright notice from the PERL Net::DNS module:
+*
+* Copyright (c) 1997-2000 Michael Fuhr.  All rights reserved.  This
+* program is free software; you can redistribute it and/or modify it
+* under the same terms as Perl itself.
+*
+* The majority of this is _NOT_ my code.  I simply ported it from the
+* PERL Net::DNS module.
+*
+* The author of the Net::DNS module is Michael Fuhr <mike@fuhr.org>
+* http://www.fuhr.org/~mfuhr/perldns/
+*
+* Michael Fuhr has nothing to with the porting of this code to PHP.
+* Any questions directly related to this class library should be directed
+* to the maintainer.
+*
+*
+* PHP versions 4 and 5
+*
+* LICENSE: This source file is subject to version 3.01 of the PHP license
+* that is available through the world-wide-web at the following URI:
+* http://www.php.net/license/3_01.txt.  If you did not receive a copy of
+* the PHP License and are unable to obtain it through the web, please
+* send a note to license@php.net so we can mail you a copy immediately.
+*
+* @category   Net
+* @package    Net_IPv4
+* @author     Eric Kilfoil <edk@ypass.net>
+* @author     Marco Kaiser <bate@php.net>
+* @author     Florian Anderiasch <fa@php.net>
+* @copyright  The PHP Group
+* @license    http://www.php.net/license/3_01.txt  PHP License 3.01
+* @version    CVS: $Id: DNS.php 302014 2010-08-09 06:17:29Z bate $
+* @link       http://pear.php.net/package/Net_DNS
+*/
+
+/* Include information {{{ */
+
+    require_once("Net/DNS/Header.php");
+    require_once("Net/DNS/Question.php");
+    require_once("Net/DNS/Packet.php");
+    require_once("Net/DNS/Resolver.php");
+    require_once("Net/DNS/RR.php");
+
+/* }}} */
+/* GLOBAL VARIABLE definitions {{{ */
+
+// Used by the Net_DNS_Resolver object to generate an ID
+
+$GLOBALS['_Net_DNS_packet_id'] = mt_rand(0, 65535);
+
+/* }}} */
+/* Net_DNS object definition (incomplete) {{{ */
+/**
+ * Initializes a resolver object
+ *
+ * Net_DNS allows you to query a nameserver for DNS  lookups.  It bypasses the
+ * system resolver library  entirely, which allows you to query any nameserver,
+ * set your own values for retries, timeouts, recursion,  etc.
+ *
+ * @author Eric Kilfoil <eric@ypass.net>
+ * @package Net_DNS
+ * @version 0.01alpha
+ */
+class Net_DNS
+{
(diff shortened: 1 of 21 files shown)


onli:
That last version is from 2010 :/ it should not be in the core. Is it even compatible with PHP7? Even if it is, it will break later on.

Net_DNS2 is linked as successor project, has a recent commit and claims better performance. Since this is for one single plugin, we'd rather merge that and change the plugin where necessary.

th-h:
> Net_DNS2 is linked as successor project, has a recent commit and claims better performance. Since this is for one single plugin, we'd rather merge that and change the plugin where necessary.

Net_DNS2 does not work out of the box with that plugin. In DNSRBL.php, line 226, it will query all blacklists and check if the host is listed:

    $response = $resolver->query($this->getHostForLookup($host, $blacklist));
            if ($response) {
                $isListed = true;
                ...

Net_DNS2 will throw an exception if the lookup fails, though, leading to a fatal error:  `Uncaught exception 'Net_DNS2_Exception' with message 'DNS request failed: The domain name referenced in the query does not exist.' `

Perhaps an `easy fix`, but not easy enough for me. :)

onli:
You can catch the exception:

```
$response = null;
try {
    $response = $resolver->query($this->getHostForLookup($host, $blacklist));
} catch (Exception $e) {  # though it would be better only to catch the Net_DNS2_Exception
    $response = null;
}
```

If it does not work that way I can have a closer look.

th-h:
Okay, works for me, see #497 

I did not check what happens when you submit a hostname instead of an IP address; there is another call to Net_DNS2_Resolver in lines 295-298. And who knows what will happen with ipv6 addresses?

So, maybe it's not perfect, but it makes the code work again.

onli:
That's good enough for now :) I'll close here and merge the PR #497. I assume you will push the fixed plugin later?

th-h:
No need to fix the plugin; the fix is in Net_DNSRBL (diverging from upstream, but I don't that's a problem).

onli:
Ah, I missed that. Diverging from upstream regularly becomes a problem - we try to update bundled_libs on every release, meaning that patch would have to be re-applied each time.

One can catch the exception everywhere up the chain, also in the plugin, which would be the right place. Then the modification to the lib could be removed. The new plugin version could get s9y 2.2 as requirement.

Doesn't have to be now and it doesn't have to be you fixing that, but we should have an issue as reminder. If you don't want to tackle this now, could you open one?

th-h:
> Ah, I missed that. Diverging from upstream regularly becomes a problem - we try to update bundled_libs on every release, meaning that patch would have to be re-applied each time.

Oh, I see. (And I should have mentioned the change in the commit message.)

> One can catch the exception everywhere up the chain, also in the plugin, which would be the right place. Then the modification to the lib could be removed. The new plugin version could get s9y 2.2 as requirement.

Unfortunately the lib has to be patched anyway to require (and use) NetDNS2 instead of NetDNS.

So we'll have to upstream the change or re-apply the patch for every release. :-(

> Doesn't have to be now and it doesn't have to be you fixing that, but we should have an issue as reminder. If you don't want to tackle this now, could you open one?

Could do that - if we agree which way to go ...

onli:
I didn't really understand the situation. Sorry!

Let's sum it up, and you correct me if I'm wrong. This Net/DNS is needed for Net_DNSBL. Net/DNSBL is only used by spamblock_rbl (and spamblock_surbl?), those are broken right now. They lie in additional_plugins.

We can't patch upstream, https://pear.php.net/bugs/bug.php?id=18883 is open and abandoned. We can't just add Net/DNS, as it is not PHP7 compatible (using the wrong constructor for example). We also have to think about ipv6 compatibility.

---

I would take a step back here. If we can't bundle a current version - because the newest version is outdated - we should remove it completely. The plugins should bundle what they need. It should be possible to rewrite them without Net_DNSBL, they can do the DNS check manually using only Net/DNS2. Or they could bring the patched Net_DNSBL.

On the other hand, it is not like DNSBL will get any updates and thus the patch would need to be re-applied, and the work is already done, and you report the plugins to be working now(right?). In that case I would not revert, but just open an issue saying "Replace Net_DNSBL functionality in X and Y (the plugin names), remove Net_DNSBL from core".

th-h:
You're correct (and yes, the plugins are working now; not all code paths have been tested, but bugs could be fixed quickly as they appear), and I agree.

PR #503, opened 2017-04-13
Fix comment preview for logged-in user.

### LINKED ISSUE #487: Comment preview while being logged in
**Reproduce:** While logged in to the backend, change the pre-filled form fields in a comment form, add a URL and comment text, and hit “Preview”.

**Result:** The preview is generated properly, but the pre-filled form fields are reset to their original values, and both the URL field and the comment text are deleted in the form (not in the preview). This does *not* happen for users that are not logged in. Obviously, the expected result would be to keep the changed values, the URL value and the comment text value in the comment form as well as in the preview.

Confirmed in 2.1.1 and 2.2.0-beta1, see [forum thread](https://board.s9y.org/viewtopic.php?p=10447059) as well.

th-h:
Comment form data was unconditionally overwritten by user data. Make that conditional ...

Fixes #487

Tested on my devblog.

### DIFF
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -1238,7 +1238,7 @@ function serendipity_printEntries($entries, $extended = 0, $preview = false, $sm
                     'is_comment_moderate'   => (isset($serendipity['GET']['csuccess']) && $serendipity['GET']['csuccess'] == 'moderate' ? true: false)
                 );
 
-                if ($serendipity['serendipityAuthedUser'] === true) {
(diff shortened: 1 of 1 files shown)


onli:
Could you change the `and` to `&&`? They are similar, but [not exactly the same](https://secure.php.net/manual/en/language.operators.precedence.php#117390), and our code normally uses `&&`.

Seems like the right fix otherwise.

th-h:
Of course!

onli:
Thanks!

Issue #512, opened 2017-05-13
"Content Encoding Error" after posting comments


th-h:
See https://board.s9y.org/viewtopic.php?f=10&t=21200 (in German)
s9y 2.1.1, PHP 5.4.45-0+deb7u8, FF 53.0.2 (seems to be browser-independent)

After posting a comment, I get the following error message from my browser (and I can't view the entry that has been commented in any way, always getting that message):

    Content Encoding Error The page you are trying to view cannot be shown because it uses an invalid or unsupported form of compression.

After some time, the error message will disappear. If I log into the backend or log out from it, the error will disappear, too.

Older comments with mandatory moderation don't demonstrate this behaviour - as long as I'm not logged in. When logged in, the comment is not moderated and will appear immediately. That seems to trigger the bug.

*Mari* on the forums has traced the problem to `serendipity_event_commentedit`, see https://board.s9y.org/viewtopic.php?f=10&t=21223#p10447271 (in German). Disabling the plugin makes the error go away. This fits in with my observations: after 300 seconds (default), comment editing is disabled, so the bug disappears after some time. And logging in or out, I guess, will reset some cookie, so the plugin doesn't recognize me as the comment author and won't let me edit the comment, so the bug disappears, too.

Unfortunately I could **not** reproduce the problem on a (slim) test installation just with `serendipity_event_commentedit` installed. There seems to be an interdependency between two plugins.

@onli, you're listed as author of `serendipity_event_commentedit`. Perhaps you have any idea?

onli:
I think I know what the issue was. The plugin echoes some javascript above the html tag, the code from https://github.com/s9y/additional_plugins/blob/12aaee2e8fdf38d15ff8e97d82286603feae4239/serendipity_event_commentedit/serendipity_event_commentedit.php#L193. This does not lead to issues in my testblog, but I can assume that it would with some charset configuration.

The easiest solution here would be to either use a different hook - `frontend_display` called in `serendipity_printComments` should have access to the comment. Alternative, save in the `fetchcomments` which ids are relevant and output them sometime later, but that might be a bad idea.

onli:
I changed commentedit to not do this anymore, in https://github.com/s9y/additional_plugins/pull/172, as it also caused issues with the new jQuery version I was testing. Given the time passed I guess it's unreasonable to ask for a test that the new version fixes this, so let's assume it does and re-open here otherwise :)

Issue #510, opened 2017-05-07
Drag'n'drop target for plugins doesn't work


yellowled:
**Reproduce:** In either event or sidebar plugins, drag a plugin from the lower end of the list to the inactive column or other sidebar.

**Expected behaviour:** Plugin is moved to the new column.

**Actual result:** Plugin is not moved to the column *unless* it is dragged to the *top* of the new column.

The drag'n'drop script is a homegrown function `getDragdropConfiguration` which I think @onli wrote? Anyway, this seems to use `.pluginmanager_container` for height calculations, which now does not *have* a height due to equal heights now being “generated” using flexbox.

If possible, we'd need to use `.pluginmanager_side` (the parent container) for calculations, altough that also contains a heading …

onli:
I remember making this work for 2.0 (but I don't remember anymore how much of it comes from the script that was used before). 

I can't reproduce this. In FF, when I drag an item from the right sidebar to another that works, in current git master. Also, looking at the script, height calculation is only used to set the height and width of the dragged item, and does not use the container for that.

There might be an incompatibility with yout used browser (chrome?) and the jquery.sortable.js plugin we use?

th-h:
Ist dies *not* work for me (FF, too) in 2.1.1. Will try to test git master on weekend.
-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

yellowled:
Don't have time to test this further atm, but yes, I used Chrome and did not test any other browsers so far.

yellowled:
Just tested in

* Chrome 58/macOS
* Firefox 53/macOS
* Safari 10.1/macOS

Maybe it's not clear enough, but you have to be dragging from the end of a plugin column that is longer than the one you're dragging to – i.e. longer than the screen so that you can not see the backend header etc. any more. Then drag to the end of the new column (just drag it across without scrolling up). That used to work before, but now it does not in any of the browsers I tested.

onli:
Ah. Okay. That I can confirm.

For the event plugin list. In the sidebar plugin list, I just have to go higher than one plugin item height. But yes, to switch columns in the event plugin list (which is a lot higher in my case) I have to scroll up all to the top.

[...]

PR #513, opened 2017-05-20
Cherry-pick bug fixes from master for 2.1.x branch.

th-h:
We fixed some very annoying bugs in master:

* Logged-in users will delete their comments when previewing them.
* Akismet in `serendipity_event_spamblock` doesn't work.
* `serendipity_event_spamblock_rbl` and `serendipity_event_spamblock_surbl` will emit fatal errors (see [here](https://www.hostblogger.de/blog/archives/6844-S9Y-2.0-auf-2.1.1-2.html))

Those fixes should go into the next patch release.

So I cherry-picked them from master to the 2.1 branch (and took another fix to suppress empty plugin groups, too).

---

We had a bit of a discussion about that at https://board.s9y.org/viewtopic.php?f=10&t=21190&p=10447234#p10447230 (nothing conclusive).

@garvinhicking @onli I kindly request a review, and if some more discussion is necessary, we could perhaps do that on the forums?


garvinhicking:
Looks very good. I'll try to join the boards this week to discuss this, but I'm a bit short on time the next few days. I agree we can and should put out a 2.1 maintenance release soon.

th-h:
Thanks!

I don't think we need to rush, but a maintenance release in the next weeks before the end of June would be very nice.

If we manage to fix some more bugs, like #512 or #510, that would be even better, but I'm out of my depth there.

PR #517, opened 2017-06-13
add template path as first entry to template_dirs array

### LINKED ISSUE #331: Smarty include relative path with 2.0.1 behaviour changed
to need strict `./` in themes.
http://board.s9y.org/viewtopic.php?f=5&t=20321 and https://github.com/s9y/additional_themes/commit/7c2d9956de69fd768e273f44ad7a8ed4289d88d6
Old `{include file="example.tpl"}` fails, new `{include file="./example.tpl"}` works.

What might have changed (when), which could influence this behaviour in Smarty? 

It must be something regarding the template fetch "scope", like the stricter set of `serendipity_getTemplateFile()` for backend chainings we have touched. (i actually can`t see why this could influence it though)

Any ideas?

### LINKED ISSUE #516: Smarty include relative path with 2.0.1 behaviour changed / reopen #331
Almost 2 years after #331 was closed...

I updated a s9y (symlinked) installation from 1.7.8 to 2.0.5 to 2.1.1 and hit the unresolved bug #331 in blogs using the kinetic theme.

Prefixing the tpl files included tpl files with ./ solves it. But digging deeper through enabling _$serendipity['smarty']->testInstall();_ in functions_smarty.inc.php I get

```
    [template_dir:Smarty:private] => Array
        (
            [0] => /var/www.serendipity/html/jugendkurse/templates/default/
            [1] => /var/www.serendipity/html/jugendkurse/templates/2k11/
            [2] => /var/www.serendipity/html/jugendkurse/templates/2k11/
            [3] => /var/www.serendipity/html/jugendkurse/plugins/
            [4] => /var/www.serendipity/html/jugendkurse/templates/default/
        )
```

which does not include the template path.
Looking at serendipity_smarty_class.inc.php (from version 2.1.1) shows that the path _$template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template'];_ is only included when _$template_engine_ is not definded. However it is set to "default".
If I patch serendipity_smarty_class.inc.php so that the templatePath is always included all the kintetic based blogs work again without modifying the tpl files. Non kinetic blogs work too.

As I have no knowledge of s9y interna my question is: is this patch a bad hack?
Why is the template path not included, which in my understanding would make sense.
Or, why is template_engine set to "default" and no the actual template. If it would be set to the template name, it would included in the template_dirs.
Please could you shed some light on this behaviour.
Thanlk you in advance.

```
--- serendipity_smarty_class.inc.php.org	2017-04-09 09:35:14.000000000 +0200
+++ serendipity_smarty_class.inc.php.igo	2017-06-09 15:26:09.922142184 +0200
@@ -131,6 +131,7 @@
             // this is tested without need actually, but it makes the directory setter fallback chain a little more precise
             $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template'];
         }
+        $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template'];
         $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['defaultTemplate'];
         $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template_backend'];

```

```
    [template_dir:Smarty:private] => Array
        (
            [0] => /var/www.serendipity/html/jugendkurse/templates/default/
            [1] => /var/www.serendipity/html/jugendkurse/templates/kinetic_jugendkurse2/
            [2] => /var/www.serendipity/html/jugendkurse/templates/2k11/
            [3] => /var/www.serendipity/html/jugendkurse/templates/2k11/
            [4] => /var/www.serendipity/html/jugendkurse/plugins/
            [5] => /var/www.serendipity/html/jugendkurse/templates/default/
        )`

```

igoltz:
add template path as first entry to template_dirs array
This should fix #331 and fix #516 without breaking other templates.
Tested with fresh 2k11 (template_engine not set) and old kinetic based blogs (template_engine set to default)

### DIFF
--- a/include/serendipity_smarty_class.inc.php
+++ b/include/serendipity_smarty_class.inc.php
@@ -121,15 +121,15 @@ class Serendipity_Smarty extends Smarty
 
         $template_engine = serendipity_get_config_var('template_engine');
         $template_dirs   = array();
-
+        
+        // first add template path
+        $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template'];
+        // then fallback engines (which should never be a comma separated list)
         if ($template_engine) {
             $p = explode(',', $template_engine);
             foreach($p AS $te) {
                 $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . trim($te) . '/';
             }
-        } else {
-            // this is tested without need actually, but it makes the directory setter fallback chain a little more precise
-            $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template'];
         }
         $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['defaultTemplate'];
         $template_dirs[] = $serendipity['serendipityPath'] . $serendipity['templatePath'] . $serendipity['template_backend'];


onli:
Thanks! 

@yellowled, @garvinhicking, that's something to keep in mind if problems related to finding templates occur during the alpha/beta.

PR #520, opened 2017-07-09
Display ChangeLog in plugin lists (if available).

th-h:
Code was using wrong file path variable for checking the presence of a ChangeLog file.


onli:
That changed a few times over the last versions, let's have an eye on this. And thanks for the PR.

PR #526, opened 2017-08-07
Fix display of messages for comment editing.

### LINKED ISSUE #525: Broken message when deleting multiple comments
The content of the message when deleting multiple comments at once is not broken into seperate lines.

See https://board.s9y.org/viewtopic.php?p=10447704#p10447704

th-h:
All messages and errors were appended to $msg and $errormsg, respectively, creating one long unformatted string - mostly unreadable.

So we make $msg and $errormsg arrays instead and iterate over those arrays in the template, displaying each message separately.

Fixes #525.

### DIFF
--- a/include/admin/comments.inc.php
+++ b/include/admin/comments.inc.php
@@ -9,23 +9,23 @@ $data = array();
 $commentsPerPage = (int)(!empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10);
 $summaryLength = 200;
 
-$errormsg = '';
-$msg = '';
+$errormsg = array();
+$msg = array();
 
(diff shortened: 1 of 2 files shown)


th-h:
It (seems to) work(s) on my testblog, but I didn't test it very thoroughly ...

include/admin/comments.inc.php
```diff
@@ -9,23 +9,23 @@
 $commentsPerPage = (int)(!empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10);
 $summaryLength = 200;
 
-$errormsg = '';
-$msg = '';
+array ($errormsg);
+array ($msg);
```
onli: I would write these lines as:

    $errormsg = array();
    $msg = array();

That way it is clear that we create a new and empty array here. array($var) reads like casting an existing variable to be an array.

th-h:
Yes, that's better. Thanks!

PR #533, opened 2017-09-21
Add xml inc svg as 'text' file extensions to serendipity_FTPChecksum()

erAck:
If accepted it would make sense to cherry-pick also to branch 2.1


onli:
Seems sound to me, but I'll defer to @garvinhicking, I'm not sure whether a different script would need to be adapted as well.

garvinhicking:
Thanks a lot for contributing, I think/hope that should be all :-)

PR #539, opened 2017-11-26
Changed behavior between body and extended part of article for nl2p

stephanbrunker:
The body part now has never a whiteline at the bottom, to keep the overview clean and the extended part has one at the top if it begins with a newline or the body ends with a newline.

### DIFF
--- a/plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
+++ b/plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
@@ -269,16 +269,23 @@ class serendipity_event_nl2br extends serendipity_event
 
                             $element = $temp['element'];
                             if ($p_tags) {
+                                // NL2P OPERATION
+                                $this->isolationtags = $isolate;
+                                
                                 $text = $eventData[$element];
                                 if (!empty($text)) {
                                     //Standardize line endings to Unix
                                     $text = str_replace(array("\r\n", "\r"), "\n", $text);
-                                    if ($element == 'body' && isset($eventData['extended'])) {  
-                                        //move newlines from body to extended
+
+                                    //move newlines from body to extended
+                                    if ($element == 'body' && isset($eventData['extended'])) 
+                                    {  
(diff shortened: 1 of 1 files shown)


onli:
Hi. Wow, I did not expect anyone tackling this code. Was hard to get in a working state. I'd be happy if you succeeded to improve upon this, and on a first glance this looks good. How did you test it?

stephanbrunker:
I use the new variant in my own blog and so far, it works fine on all the articles. Especially those containing images (lots of div, a, and img tags), tables, lists and blockquotes. That is the testing in a production environment. I made it so that the output html is human-readable, so you can take a look at the html source of www.intjblog.de .

Then I have the core online on a test php server: http://s521456705.online.de/webspace/testnl2br.php 
This page has only two textareas, in the first you put your article, the second one outputs then the formatted result. I tested a lot of possible constellations there. One drawback might be that it only accepts the html tags which I've defined, but I took them from the html5 spec and such tags like meta shouldn't be in an article anyway. Everthing else is escaped to &lt; &gt; . 

The article gets crawled from front to back, after being exploded along tags. Inside the text blocks defined by the block tags (even nested ones like div or table) the content gets processed according to the tag: if p is allowed then the next stage is executed, otherwise the text is unprocessed. The second stage too builds up a stack of active (in this case, the inline allowed ones) tags, so if you insert a newline, the active tags are closed, the paragraph is inserted, next paragraph is opened and the active tags restored. 

That should cover everything, but if you find a bug, I can quickly remove it.

stephanbrunker:
test:
`<i>aaa<b>bbb</i>ccc</b>ddd` 
is converted to:
`<p class="break"><i>aaa<b>bbb</i>ccc</b>ddd</i></p>`

means that the first (illegal) closing tag is ignored and the tag is then closed correctly at the end of the paragraph.

onli:
Good enough for me. Thank you very much for your work! This will go into the current master and thus be tested in the coming beta circle. I'm really looking forward to see how this one works out in practice :)

PR #540, opened 2017-12-19
added iframe to allowed singleton tags

stephanbrunker:
Found a small bug: iframe tag was not in the allowed tags, so an embedded YouTube video wouldn't work.


onli:
Thanks!

PR #541, opened 2017-12-27
allowed nested lists, svg XML elements and fixed bugs in tag_clean

stephanbrunker:
i've made further improvements in the nl2p functionality, like nested lists, allowing svg tags (which means the xml elements inside too) and there was a bug in the tag_cleaner. I've also a new fork (not this commit) for this plugin where I changed the nl2p that double newlines make paragraphs and single newlines br-tags. That means that the CSS for the paragraphs is unmodified by the plugin (only the margin=0 cases), so the paragraphs have the default margins. What also should be done is the description of the isolating tags. You shouldn't use code,pre,textarea as isolating tags, because that is done automatically now. If you want to isolate manually, you should use the nl-tag. But I can only change that for english and german, not all the other languages. And if the new mixed newline to br and p functionality is to be integrated, that would mean either replace the existing or adding a new option. Don't know what would be best ... programmatically, there are only other constants for the paragraphs and a different second stage (after splitting it into block elements). I saw the necessity for that if you use headline tags and you don't want the distance between the headline and the paragraph with a fixed 1em, because the headlines already have a margin-bottom. With the new fork, these settings are based on the existing CSS.

### DIFF
--- a/plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
+++ b/plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
@@ -400,31 +400,42 @@ p.break {
 	 */
 
 	// following w3.org, these elements close p elements automatically:
-	var $block_elements = array('table','ul','ol','pre', 'dl',
+	var $block_elements = array('table', 
 								'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
-								'menu', 'section',
+								'menu', 'section', 
 								'address', 'article', 'aside', 'fieldset', 'footer',
 								'form', 'header', 'hgroup', 'hr', 'main', 'nav', 'p'
 								);
 
-	var $nested_block_elements = array('div','table','blockquote');
+	var $nested_block_elements = array('div','table','blockquote','ul','ol','dl');
 	
-	var $singleton_elements = array('area', 'br', 'col', 'command', 'embed', 'hr', 
+	var $singleton_block_elements = array('hr');
+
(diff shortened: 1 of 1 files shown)


onli:
Hi, I'm gonna merge this into master now. 

For the idea to mix p and br tags: I think that is a good idea. It has the advantage of working when visiting the page directly and when reading the article via the feed (in a feedreader or a planet). On the other hand, I used p-tags here because I wanted to make clear that semantically there are different kinds of paragraphs - but br tags do not really mean that the following is a paragraph. What do you think?

I would not want it as an option (nl2p mixed into nl2br is already complicated enough), but we could just change it, if you feel that being more reliable / needing no additional css is more important than the semantics.

stephanbrunker:
My idea was that dividing the content into paragraphs depending on double line breaks and br-tags for single line breaks is semantically the correct way. I don't know how WISYWYG-Editors handle that, but there you can also insert a single or double line break. So, if you have a clearly separated paragraph in your raw input, you have one in the output, and if you make only a simple break (for example inside a table cell for better readability, you get that without starting a new paragraph. Also, the output is much cleaner that way. I'm actually testing this version and will make a PR then if it's ready.

Funnily, that brings back the behaviour because of I originally started to mess with the plugin: The article body part then comes with a margin-bottom or without, which makes the overview a bit unstable to look at if the devision between body and extended is not intended with a margin (the natural styling for block elements changes only the margin-bottom for h1 ... h6 and p elements) But that is easily to fix with a user-CSS .read-more { margin-top 1em; }

PR #542, opened 2018-02-12
Rework autologin to use a token approach

onli:
The prior code stored encrypted user data in the cookie that was then checked. This new approach is cleaner, as it only stores a token, and it does not use problematic crypto functions deprecated in PHP 7.2


onli:
I'l merge this already, for testing during the camp

PR #543, opened 2018-03-13
Move from SHA1 to bcrypt for stored passwords

onli:
SHA1 is not an ideal password hash, even when salted, because it is cheap to compute. Since version 5.5 PHP offers bcrypt built in, which is a more expensive and secure hash function well suited for passwords.

This will need more testing, but it builds on what worked when we moved away from md5, so I think there is a good chance this will work fine.

### DIFF
--- a/include/functions_config.inc.php
+++ b/include/functions_config.inc.php
@@ -549,9 +549,8 @@ function serendipity_authenticate_author($username = '', $password = '', $is_has
                 if ($is_valid_user) continue;
                 $is_valid_user = false;
 
-                // Old MD5 hashing routine. Will convert user.
                 if (empty($row['hashtype']) || $row['hashtype'] == 0) {
-
+                    // Old MD5 hashing routine. Will convert user.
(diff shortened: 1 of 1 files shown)


garvinhicking:
@onli This looks nice generally. However currently it looks to me that the bcrypt style does not use a salt? We definitely should use a installation-unique hashkey, so if the serendipity_authors database table is leaked, all passwords are not easily matchable with a brute force dictionary attack. The current sha1 implementation uses this hashkey.

(What I mean is that for example if someone knows that "rainbow" is bcrypted as "1237sdfb", when they see "1237sdfb" inside a database dump, they shall not directly know that those users had "rainbow" as a password". With a salt, the actually stored password would be "SALTrainbow" and so it will not resolve to "1237sdfb" instead. I am not a great authority on encryption, but for sha1 this all sounded sound to me ;-) )

onli:
Bcrypt automatically use a salt for each password, there is no need to that manually :)

onli:
I'l merge this already, for testing during the camp

PR #544, opened 2018-03-22
Add functions for entry check

guillaumezin:
This modifications have to do with another PR https://github.com/s9y/additional_plugins/pull/72 for serendipity_event_entrycheck, cf. bug https://board.s9y.org/viewtopic.php?f=3&t=23375&sid=9d554ed2b086342e2d3a72ab766b1a02


onli:
Hi. Thanks a lot for sending a PR! I'm gonna merge this. Small thing is that it would be nicer to use an event listener, instead of using an onclick handler in the html. That would go into the `/templates/2k11/admin/serendipity_editor.js.tpl`, maybe you are interested to move that over?

PR #557, opened 2018-07-01
Improved nl2br

stephanbrunker:
Because of a lack of time, it took a while, but I had an improvement to make which extends the one I did end of last year. Basically, the plugin uses now a much cleaner html with simple <p> </p> tags. Only where no margin top and bottom is wanted (that means between body and extended article), special classes are used. The second improvement is that single newlines convert to <br> tags. Additionally, i made a lot of fixes in the tag_clean and explode_along_tags function so that all non-valid < and > characters are escaped. Also the newline converting is only made where the paragraphs are allowed by html. The basic functionality is unchanged so existing articles don't have to be changed. Simple put: you can enter simple text, and the plugin does the formatting for you, or you spice it up with as many html as you want and the plugin only works in the unformatted parts.


onli:
Okay. I'm looking forward to see this in action, thus I merged it right now to increase the chances this gets tested a lot. Like last time: Thanks a lot :)

stephanbrunker:
I used it in my own blog first and saw no obvious issues. But if something rises up, I can fix that quickly. Normally, users think in ways unimaginable to the developers ....

onli:
Hi @stephanbrunker, there seem to be some issues in this code. I'm testing the current git master in my blog right now and have a look at https://www.onli-blogging.de/1496/Prismatik-unter-Ubuntu-mit-Hyperion-ersetzen.html: all the code blocks do not have my <pre class="code"> anymore wrapping them, they got replaced by <p>s. I confirmed that nl2br is the issue here. 

I also do not like that there now is always whitespace between the last paragraph and the footer, that would have been a <p class="linebreak"> before and not introduce that whitespace.

Could you have a look at this and try to fix it?

stephanbrunker:
Hello onli, 

your mail was somehow clipped. But I think I get the essentials.

The whitespace: My reasoning was that all the paragraphs are paragraphs and should be formatted as such, means: how they are done in the CSS. I only made the exception that if there is no line breaks between body and extended part of the article that there are extra classes introduced with margin-top 0em and margin-bottom 0em. So, if your body ends without break and your extended article starts without break, there should be no margin between both parts. The simple line break can’t be avoided because these are two different div-elements, which are block elements per definition.

This withspace at the end of body and extended is no line break, it is simply the margin of the p-element and can be altered in the user.css of the template, for example with

.serendipity_entry_body p:last-of-type { margin-bottom: 0em; } /* that is the margin after the body part 
#extended p:last-of-type { margin-bottom: 0em; }               /* that is the margin after the extended part

And so on. In that case, it also depends of the style of the following a-element, which is set in the template and it can have an margin-top.

Preformatted code:
It should be that preformatted code: 
<pre><code> 
line1
line2
line3
</code></pre>

should be displayed unaltered. I also found the problem. The plugin had (not my idea) the option to use a user-defined tag (something like <nl> ) for isolation and that one should be purged (because it is not standard html) in the output. I threw this and the html pre and textarea tags in one bucket so the breaks are in the output source but the pre-tag were purged, too. That one was simple to fix. But to get the output right, you have to have the code inside the <pre> tags, that is the semantically correct way to do it. 

I also updated the english and german description of the plugin, so that it is clear what the p-tag version does and that everything inside pre, svg and style tags is automatically ignored and not handled by the plugin.

I’ll make a pull request for the updated files.

Regards,
Stephan

onli:
Hi
I had html tags in my first version of this comment and did not realize Github would not encode it.

The logic of the nl2p mode was: 

 1. If a block of text is followed by one \nl, wrap it in `p class="linebreak"`
 2. If a block of text is followed by two or more \nl, wrap it in  `p class="whitespace"`

I'll first test your patch now (thanks!) and think about the smaller issue later, though I really see a problem with the current situation, as now an update to the alpha changes how blog articles look. Maybe we should have the plugin emit the CSS to fix this and then it's okay, but the pre tag is more important so far.

stephanbrunker:
I thought about how the plugin should work, with the intent to produce a well-written html out of plain text with some html in it. And my solution was: Change simple line breaks into br-tags and double breaks into p-tags. I think in the first run, I only improved the existing logic, but after I added more and more intelligence to it, it seemed like a dead end regarding the possibilities which elements are valid parents for p-tags and so on. Then I dropped these classes and changed it. But it shouldn't change the general look of an article: there is still only a line break or a 1em margin. The only conflicting point is the last line of the article and there it is more a matter of personal opinion and that should be addressed by the author in the CSS or by the template, because that much depends on which elements are next below the article.

onli:
> The only conflicting point is the last line of the article

Right, apart from that it works really well :)

>  that should be addressed by the author in the CSS or by the template, because that much depends on which elements are next below the article.

I think it should be fixed in the plugin, but we can fix it with CSS emitted by the plugin. The themes are already built without that additional margin in mind, for entries and comments. A user would have a blog running as he wants, update to the new s9y version and then have to change its theme to react to this change in nl2br. I'll have a try with the CSS, your idea above with `.serendipity_entry_body p:last-of-type` could work fine.

onli:
Hi. I found a new related bug. Please have a look at https://www.onli-blogging.de/242/Toshiba-Satellite-4030-CDT-mit-Linux.html. The article looks fine just until first the &lt;strike>. Afterwards all the text is transformed to be all lower case and the html is not interpreted anymore. In my tests it looks like the nl2br plugin is responsible here, deactivating it fixes the issue. Do you have an idea where the issue is exactly?

stephanbrunker:
Because strike is not recommended in HTML5, I didn't add it to the list of inline-tags. That one is easy to fix. The bug is that in this case, the plugin should only escape the tags of the strike and not everything else. I'm going to take a look at the escaping feature, it should reset if a valid tag occurs. The idea is that you can use the &lt; &gt; without the need to escape them (like in smaller or bigger as), to do it automatically if they aren't part of a valid html tag.

PR #565, opened 2018-07-31
replace deprecated each() with foreach()

### LINKED ISSUE #562: Use of deprecated "each()" function
On a test installation with PHP 7.2 I got this stacktrace:
```
Fatal error: Uncaught ErrorException: Notice: The each() function is deprecated. This message will be suppressed on further calls in [path]/include/plugin_api.inc.php:1123
Stack trace:
#0 [internal function]: errorToExceptionHandler(8192, 'The each() func...', '/home/hanno/web...', 1123, Array)
#1 [path]/include/plugin_api.inc.php(1123): each(Array)
#2 [path]/include/functions_config.inc.php(532): serendipity_plugin_api::hook_event('backend_login', false, NULL)
#3 [path]/include/functions_config.inc.php(374): serendipity_authenticate_author('', '', false, false)
#4 [path]/serendipity_config.inc.php(371): serendipity_login(false)
#5 [path]/index.php(12): include('/home/hanno/web...')
#6 {main}
thrown in [path]/include/plugin_api.inc.php on line 1123
```

The call to "each()" is still in the latest git code:
https://github.com/s9y/Serendipity/blob/master/include/plugin_api.inc.php#L1123

According to PHP docs ( http://php.net/each ):
"This function has been DEPRECATED as of PHP 7.2.0. Relying on this function is highly discouraged."

hannob:
This fixes #562.

The "each()" function is deprecated, this is a rewrite with foreach. As foreach creates copies and not pointers we can't modify the variable directly, so we access it via its array key.

### DIFF
--- a/include/plugin_api.inc.php
+++ b/include/plugin_api.inc.php
@@ -1118,9 +1118,7 @@ class serendipity_plugin_api
         }
 
         if (is_array($plugins)) {
-            // foreach() operates on copies of values, but we want to operate on references, so we use while()
-            @reset($plugins);
-            while(list($plugin, $plugin_data) = each($plugins)) {
+            foreach($plugins as $plugin => $plugin_data) {
                 $bag    = &$plugin_data['b'];
                 $phooks = &$bag->get('event_hooks');
                 if (isset($phooks[$event_name])) {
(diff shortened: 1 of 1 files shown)


onli:
Wow, even if it were not for compatibility it is a lot cleaner like that. I'll merge this directly.

PR #570, opened 2018-08-09
Return output

### LINKED ISSUE #536: Missing output of captcha
The spamblock plugin does not show the captcha image in my s9y 2.1.1 installation.
 I added "return $output;" as a new line at 815 in the file "plugins/serendipity_event_spamblock/serendipity_event_spamblock.php" and it works now.
Please validate this fix and if ok take on for the next version.

titanofold:
When $use_gd is false, returning $output so that the captcha images that have been generated are actually embedded into the final rendering of the page.

Fixes #536.


onli:
Thanks!

PR #580, opened 2018-09-25
event_spamblock: Fixed forced comment moderation 0 -> moderate all

stephanbrunker:
for the option: "Moderate comments after x days" the function was missing that for a value of 0, all comments are going to be moderated. I fixed this.

### DIFF
--- a/include/admin/upgrader.inc.php
+++ b/include/admin/upgrader.inc.php
@@ -270,6 +270,11 @@ $tasks = array(array('version'   => '0.5.1',
                      'function'  => 'serendipity_installFiles',
                      'title'     => 'Update .htaccess file',
                      'desc'      => 'Adds a new "documentation.*.html" rewrite rule exception to allow calling plugin documentation URLs.'),
+
+               array('version'   => '2.2.1-alpha2',
+                     'function'  => 'serendipity_upgrader_spamblock_moveForce',
+                     'title'     => 'Swap value forcemoderation',
+                     'desc'      => 'If force moderation of comments is activated, activate the new option moderation_auto'),
 );
 
 /* Fetch SQL files which needs to be run */
--- a/include/functions_upgrader.inc.php
+++ b/include/functions_upgrader.inc.php
@@ -423,9 +423,10 @@ function serendipity_upgrader_spamblock_moveForce() {
     global $serendipity;
 
     $force = serendipity_db_query("SELECT value FROM {$serendipity['dbPrefix']}config WHERE name LIKE '%serendipity_event_spamblock%forcemoderation'", true);
+    if (is_array($force)) { $force = $force[0]; }
     if ($force > 0) {
          serendipity_db_query("UPDATE {$serendipity['dbPrefix']}config SET value = 'true' WHERE name LIKE '%serendipity_event_spamblock%moderation_auto'" );
-     }
+    }
 }
 
 function serendipity_upgrader_move_syndication_config() {
--- a/include/functions_upgrader.inc.php
+++ b/include/functions_upgrader.inc.php
@@ -419,6 +419,15 @@ function serendipity_upgrader_rewriteFeedIcon() {
     serendipity_db_query("UPDATE {$serendipity['dbPrefix']}config SET value = '" . serendipity_db_escape_string($path) . "' WHERE name LIKE '%serendipity_plugin_syndication%big_img'");
 }
 
+function serendipity_upgrader_spamblock_moveForce() {
+    global $serendipity;
+
+    $force = serendipity_db_query("SELECT value FROM {$serendipity['dbPrefix']}config WHERE name LIKE '%serendipity_event_spamblock%forcemoderation'", true);
+    if ($force > 0) {
+         serendipity_db_query("UPDATE {$serendipity['dbPrefix']}config SET value = 'true' WHERE name LIKE '%serendipity_event_spamblock%moderation_auto'" );
+     }
+}
+
 function serendipity_upgrader_move_syndication_config() {
     global $serendipity;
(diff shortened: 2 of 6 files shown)


onli:
The code is explicitly doing the contrary: 0 means no moderation. That also matches the (german) description:

> 0 bedeutet, dass keine automatische Moderation erzeugt wird.

We often need a way to have a null value that disables an option.

When wanting to moderate all comments this can be set in the personal settings as a default setting for the entry editor.

Is this maybe more about wanting a way to force-moderate all comments, of new and existing entries?

stephanbrunker:
Oh … if 0 means no moderation, then that setting is not consistent to the captcha setting slightly above where 0 means "force captchas". And yes, the goal is to force-moderate all existing entries. If the value is 0.0000116, this means after 1 second, so logically 0 means immediately. To turn the auto-moderation off, a value of "infinite" would be correct, so perhaps a -1 ? Or an extra option to turn it off, like it is for the captchas?

onli:
Right, that is not consistent at all.

> Or an extra option to turn it off, like it is for the captchas?

That sounds best to me. Then it would be like the captcha setting. *Auto-moderate* comments yes/no, then the time until auto-moderation, default being 0.

Maybe you can add this to this PR? The default should be off.

But then we should also have an upgrader task, that sets the new boolean *Auto-moderate* option to false if the interval is 0 now, and to true if it is a higher number.

stephanbrunker:
I can do parts of this task, like the changes in the event_spamblock.php and the DE and EN language files. What I cannot do are the other languages and the upgrader/database operations. 

This means I add 'forcemoderation_auto' into introspect->propbag ('configuration') and a new case at introspect_config_item? Is the field in the config database then created automatically with this 'propbag' or is that something which has to be done manually first?

stephanbrunker:
I did what I can do - the event_spamblock.php and EN and DE language files.

stephanbrunker:
This is much like poking in the mist, but I think it could be mostly correct. I am not sure about the version number, but the function should activate the moderation if a non-zero value is set. If it is set to zero and thus the moderation is off, the default value 'false' doesn't have to be changed.

onli:
The logic looks fine, thank you. Did you test the function itself? I'm surprised that `serendipity_db_query` would return a straight value and not an array.

stephanbrunker:
No, as I wrote I honestly don't have a clue of what I coded (outside of SQL and PHP in general). I just made it up by altering one of the existing functions and looked up the names of the db fields. I have an active installation of serendipity, but not a test environment, so I couldn't test it. (I also don't know which version number I have to insert in the upgrader.inc.php) If serendipity_db_query returns an array, then I have to copy&paste that snippet with if(is_array()) ... , too. I presumed it would return only a value because there is only one value matching the query. I only worked with mysqli->query, num_rows and fetch_assoc before, not with the serendipity implementation of the database handling. Because all other SELECT calls with serendipity_db_query use is_array, I changed that now.

onli:
Okay. So the next step is to test this, it looks like it should work fine. I have a test installation and will try to do that soon.

th-h:
With the recent commits, the PR affects not only `serendipity_event_spamblock`, but also `serendipity_event_nl2br`. I do not think that was really intended ...

stephanbrunker:
I am quite new to git. This pull request it did in autumn was still open and my working copy of the repo had these commits around the spamblock. I use this changes since then in my own blog because I have a bunch of nasty spammers blocking some older articles. Then a new request came up to bugfix nl2br in between. I created a new branch and cherry-picked that commit, and this PR was merged into the master by onli yesterday. It may be that this PR (from my master) has now also the nl2br commit into it, but that doesn't matter as these are already merged as https://github.com/s9y/Serendipity/commit/04cc9d06996906168c29d860b525d4fdb119b8da

onli:
I'm actually not sure that it does not matter, the file changes tab shows that merging this PR would change files in nl2br. Which is actually really strange if those change were already in the other PR. Would you be okay with opening a new PR with just the relevant changes? If you still don't have the resources to test the upgrader I'd then test it and merge it asap.

stephanbrunker:
That should do it. I reverted my nl2br commit, and merged and rebased with origin (where the commit was already applied). So the difference are only these four commits in spamblock and upgrader.

onli:
> My git cheatsheet is still growing ...

Well done though :) I wasn't sure how to solve this either.

The code looks very good. I will merge this first and test then, this should work. Could you add one commit more that bumps the version number of the spamblock plugin to 1.87? That way all the relevant code changes are included in this PR.

onli:
Thanks!

onli:
Okay. I tested the upgrade task now and in both configurations - active via a day limit or set to 0 to be inactive - the upgrade task worked as expected. The code works fine :)

However, it's more inconsistent now than I realized. You are right, it now works a bit like the captcha toggle. But we also have an option "Trackback/Pingbackmoderation nach wievielen Tagen erzwingen" where 0 still means deactivated. We also have the IP-Block Interval, where 0 means not to block. And "Erforderliche Anzahl an Links für Moderation" as well as "Erforderliche Anzahl an Links für Abweisung" also work with 0 being the value that deactivates that filter. I assume we have more options like that in other plugin configurations.

Basically, the only inconsistent option I see was the "Captchas nach wie vielen Tagen erzwingen" option, which is the one this PR is modeled after :/ And that option is special because the toggle above also swaps between the captcha types.

---

So I'm now thinking: Maybe we should restore how it worked before and use the code to fix the captcha setting. Or alternatively, do we maybe need a configuration type "intandbool" that creates a widget where you can enter a number or also control whether the option is being active with a checkbox? Hierarchic menu options?

PR #590, opened 2019-02-22
event_nl2br: fixed isolation issue for pre-tags. Updated description.

stephanbrunker:
I have fixed the display of code or other text which should displayed as it is. Use it with the html pre tag. There was a bug that purged the tag from the output. The newlines were copied into the output, but without the pre-tag, html doesn't display these.

### DIFF
--- a/plugins/serendipity_event_nl2br/UTF-8/lang_de.inc.php
+++ b/plugins/serendipity_event_nl2br/UTF-8/lang_de.inc.php
@@ -8,11 +8,11 @@
  */
 
 @define('PLUGIN_EVENT_NL2BR_NAME', 'Textformatierung: NL2BR');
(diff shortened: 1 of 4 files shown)


stephanbrunker:
I just saw the parts of your post which got clipped in the mail. The question is how to handle the last paragraphs, and which ones. The last of "body" or "extended"? And what if they are both displayed after another, which the plugin cannot detect? I still think that the best approach is to use standard p-tags and make the formatting in the user.css, with last-of-type if you wish it different. There is also the issue of the following element and it's margin-top, which also influences the margin between these parts.

onli:
I merged this already, because it properly fixes the pre tag issue I observed in my blog. I'll put a commit on top of it that raises the version number and adds a note to docs/NEWS.

PR #595, opened 2019-03-03
update composer.json (php 7.0) + update "voku/simple-cache"

voku:
This change is [Reviewable](https://reviewable.io/reviews/s9y/serendipity/595)


onli:
Does voku/simple-cache require PHP 7.0? I think that would be okay, but something we would need to document when releasing the new version.

voku:
yep, since version 3, I dropped support for PHP < 7.0

onli:
Thank you for the PR! I merged to test it in practice now, too big to review in detail before.

PR #599, opened 2019-03-08
fixed missing 'strike' tag and fixed tag recognition

stephanbrunker:
Fixed the issue with nl2br. It now converts to lowercase only when a html tag is detecte, added the strike to the inline elements and improved the detection … more text here than real changes.


onli:
Great, thank you! I'll merge and test it directly.

PS: It misses the version number bump ;)

onli:
Works!

PR #600, opened 2019-03-10
Make sure there is a $serendipity['lang'] defined

HQJaTu:
If for any reason `serendipity_load_configuration()` fails to load conf variable `lang` for author id 0, the fallout is pretty horrible and hard to pinpoint to original source. The emitted error of _PHP Fatal error:  Uncaught Error: Call to undefined function serendipity_mb()_ has very little clues what's really going on. That's why having a fallback in place can be valuable.

Not having `lang` is a corner case, but it can happen, especially for upgraded installations.


onli:
Seems useful, thank you!

PR #602, opened 2019-03-17
Added codeTag plugin into CKeditor

HQJaTu:
People, like me, who blog about computers will love to have monospace font in form of \<code\> and \<tt\> HTML tags. Enabling those HTML-tags and added plugin https://ckeditor.com/cke4/addon/codeTag into CKeditor.


htmlarea/ckeditor_s9y_config.js
```diff
@@ -252,7 +253,8 @@ CKEDITOR.editorConfig = function( config ) {
         { name: 'mediaembed',   items: [ 'MediaEmbed' ] },
         { name: 'others',       items: s9ypluginbuttons },
         { name: 'document',    groups: [ 'mode', 'document', 'doctools' ], items: [ 'Source' ] },
-        { name: 'about',        items: [ 'About' ] }
+        { name: 'about',        items: [ 'About' ] },
+        //{ name: 'codetag',     items: [ 'Code' ] }
```
onli: Sure you do not want to enable the codetag in that default mode?

I'd activate it, but if you'd rather not it would be better to remove that line from the PR.

onli: One thing more: tt tags? Those are not valid HTML 5. How about just enabling the code tags?

HQJaTu: Oh, I would. I definitely would love to see it enabled for everybody.

Its just that I don't know how it fits the current product vision of S9y. The feature hasn't existed for general public since 2.0 was released, me personally will need it in almost every single blog post I ever write. I do write about computers and software engineering, after all. To make up some usage statistics from the top of my head: 99,9% people won't.

If you will enable it for all, that's great! If you won't, please make sure it is easily available for anybody needing the feature.

onli: Okay.

Honestly, for the CKEditor there is not much vision. The developer that forced the main changes left  the project, I just wanted a simple enough toolbar with only commonly used buttons. None of us developers use the wysiwyg editor. I agree though that code blocks should be part of it.

Cab we bundle it in this PR? You can just push new commits onto that branch and it will show up here. It would be great if you could:

 1. Remove tt fro mthe list of allowed content, as tt is not a valid HTML5 tag anymore. Or is that a fix part of the code plugin and we can't have that without tt?
 1. Move the code block further to the left. I propose to have it above the colors section in https://github.com/s9y/Serendipity/blob/6b64e4c2c01444afb6728f06a21295e4000cd55e/htmlarea/ckeditor_s9y_config.js#L232 and above the media embed in https://github.com/s9y/Serendipity/blob/6b64e4c2c01444afb6728f06a21295e4000cd55e/htmlarea/ckeditor_s9y_config.js#L253
 1. That's included in 2, but let's activate it for the default toolbar.

HQJaTu: I'll make the suggested changes and create a new PR.

TT is there to support some of my first ever blog posts. Obviously I haven't edited those for years, so I won't be needing it.

HQJaTu:
The new commit has your changes in it.

Maybe I will need to learn how to read. You wrote earlier, that any changes will appear here. No need to create a new PR.

onli:
> No need to create a new PR. 

Easier that way :)

I did not test now how that looks and behaves in the editor itself, I'll just assume it works fine, and we or you can still change stuff later of course. But I'll merge this now :+1:

PR #614, opened 2019-07-28
Remove curly braces array syntax, avoids warning in PHP 7.4.

hannob:
Accessing arrays with a curly brace is deprecated and throws a warning in PHP 7.4:

PHP Deprecated:  Array and string offset access syntax with curly braces is deprecated in /[path]/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php on line 1323


onli:
Thanks! Really strang code, `parts[1] - 1`? But the captcha is still working?

hannob:
I can confirm that with the patch applied captchas and comments still work. Given that the change is so trivial I would also be very surprised.

I haven't tried understanding the code in question though.

onli:
If code is very strange it's better to ask once more, it might be misleading. Merged, Thanks again!

PR #619, opened 2019-08-10
[event_spartacus] Don't cache wrong plugin list.

th-h:
Declaring `$pluginlist` as _static_ will cache the database query results and always return the same list of plugins - no matter which type is queried (sidebar or event).

This cache must be busted if another type of plugins is queried.


onli:
How do you feel about using the serendipity cache functions instead of the static variable? To make the survive longer than one request cycle?

th-h:
> How do you feel about using the serendipity cache functions instead of the static variable? To make the survive longer than one request cycle?

I'm all for it, but have no idea how they work. :) I just improved on what I found ...

PR #639, opened 2019-09-17
Escape category images to avoid backend XSS

hannob:
It is possible to cause a backend XSS via the category icons.

PoC:
1. Create a category with something like "> in the "Category Image" field.
2. Start a blog post, select this category and click on Preview.

I'm aware that s9y isn't really protected against backend XSS due to the blogposts itself being not XSS safe, but still I think output should be properly escaped.

This is a fix within the templates, so naturally other templates won't automatically get that fix.

### DIFF
--- a/templates/2k11/entries.tpl
+++ b/templates/2k11/entries.tpl
@@ -10,7 +10,7 @@
         </header>
 
         <div class="clearfix content serendipity_entry_body">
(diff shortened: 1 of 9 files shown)


onli:
Does this really need a backport currently, @th-h ?

onli:
And thank you, @hannob !

PR #645, opened 2019-10-13
2.3 BBcode plugin 

FabienChab:
In the BBcode plugin, ordered lists doesn't work for list=i and list=I
Solution:
after line 300 add
`.bb-list-ordered-lr{
	list-style-type:lower-roman
}
.bb-list-ordered-ur{
	list-style-type:upper-roman
}`


th-h:
That doesn't work this way in a pull request, pushing from master into 2.3. :-)
See issue #647 instead.

Thanks for reporting!

PR #650, opened 2019-10-13
Fix a lot of single issues and overhaul the language settings

stephanbrunker:
* admin/personal.inc.php: removed setting of $_SESSION: serendipity_lang
     (already set in serendipity_config.inc.php for all scenarios)
   * compat.inc.php: function detectLanguage returns now NULL if
     the detection failed, also doesn't change $serendipity['autolang']
     anymore which remains 'en' as general fallback.
   * functions.entries.php: 
        * function serendipity_printArchives() hooks now
          into hook_event::'frontend_fetchentries' to get multilingual data, 
          also doesn't show months and years with zero entries.
        * function fetchEntryCategories, fetchEntryData and fetchCategories 
          strips now data of multilingual tags
   * plugin_api.inc.php: added function find_plugin_id which returns an 
     array of the ids of the installed instances of the given plugin
     this is used for checking the configuration of that plugin in other
     plugins
   * functions.config.inc.php: function serendipity_load_configuration
     sets the language back to 'autolang' (which is 'en') if the language
     loaded does not exist. Also sets the value $serendipity['default_lang']
     to the default language of the blog.
   * functions.config.inc.php: removed the Cookie 'userDefLang' which is no
     longer needed
   * functions.config.inc.php: major rework of the functions 
     serendipity_getSessionLanguage and serendipity_getPostAuthSessionLanguage.
     Those two functions should now process the user input via GET and POST,
     store it into SESSION and COOKIE, and if no such input is given, fallback
     to browser language negotiation and default language. More details in the
     comments to that functions. $serendipity['detected_lang'] is the language
     found from user input or browser negotiation.
   * functions.routing.inc.php: fix some lookups from $_GET to $serendipity['GET']
   * admin/plugins.inc.php: fix some lookups from $_GET to $serendipity['GET']
   * created a new event hook 'multilingual_strip_langs' to strip tagged translation of elements 
     outside the sidebar (mostly categories in entry display)

event_entryproperties:
   * removed 'WHERE' keyword from fetchentries event hook, because that should be handled 
      by the issuing SQL

event_nl2br:
   * just changed the indent to spaces for the piece of code i'd written in my last commit

event_spartacus:
   * fixed a call to a NULL value if the server cannot make a connection

plugin_categories:
   * stripped the translation tags from the links

plugin_recententries:
   * using the sql condition set aquired by event_hook::'frontend_fetchentries'
      to get the multilingual data set
   * $this->title (= title in backend) set to plugin name

plugin_superuser:
   * changed the link to /serendipity_admin.php

### DIFF
--- a/include/functions_config.inc.php
+++ b/include/functions_config.inc.php
@@ -328,8 +328,15 @@ function serendipity_load_configuration($author = null) {
         $serendipity['baseURL'] = $serendipity['defaultBaseURL'];
     }
 
-    // Store default language
+    // check language selection or fallback
+    if (isset($serendipity['lang']) && !isset($serendipity['languages'][$serendipity['lang']])) {
+        $serendipity['lang'] = $serendipity['autolang'];
+    }
+    
+    if (empty($author)) {
     $serendipity['default_lang'] = $serendipity['lang'];
+}
+
 }
 
 /**
@@ -661,7 +668,6 @@ function serendipity_load_userdata($username) {
     $_SESSION['serendipityHashType']     = $serendipity['serendipityHashType']     = $row['hashtype'];
 
     serendipity_load_configuration($serendipity['authorid']);
-    serendipity_setCookie('userDefLang', $serendipity['lang'], false);
     return true;
 }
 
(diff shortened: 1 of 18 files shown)


include/functions_entries.inc.php
```diff
@@ -146,10 +146,15 @@ function &serendipity_fetchEntryCategories($entryid) {
                 ORDER BY c.category_name ASC";
 
         $cat =& serendipity_db_query($query);
+        
         if (!is_array($cat)) {
             $arr = array();
             return $arr;
         } else {
+            foreach ($cat as $key => $catdata) {
```
garvinhicking: This is not very good performance-wise to call it for each key individually. Please move this hook outside the foreach loop to operate on the whole $cat array?

include/functions_config.inc.php
```diff
@@ -1065,29 +1094,20 @@ function serendipity_getSessionLanguage() {
...
-        $lang = $serendipity['lang'];
+        $lang = $serendipity['lang']; //$serendipity['lang'] is overwritten by load_configuration(user) and points to the user default language
     }
 
+    // save the result into SESSION
     $_SESSION['serendipityLanguage'] = $lang;
 
     if ($serendipity['expose_s9y']) serendipity_header('X-Serendipity-InterfaceLang: ' . $lang);
 
-    if ($lang != $serendipity['lang']) {
```
garvinhicking: Are there maybe plugins or other instances that actually work in $serendipity['content_lang']? It feels a bit dangerous to remove this variable completely?

include/functions_entries.inc.php
```diff
@@ -511,6 +516,8 @@ function serendipity_fetchEntryData(&$ret) {
 
     if (is_array($search_ret)) {
         foreach($search_ret AS $i => $entry) {
+            serendipity_plugin_api::hook_event('multilingual_strip_langs',$entry['category_name']);
```
garvinhicking: See above

include/functions_entries.inc.php
```diff
@@ -722,9 +729,16 @@ function &serendipity_fetchCategories($authorid = null, $name = null, $order = n
           $flat_cats = array();
           $flat_cats[0] = NO_CATEGORY;
           foreach($cats AS $catidx => $catdata) {
+              serendipity_plugin_api::hook_event('multilingual_strip_langs', $catdata['category_name']);
               $flat_cats[$catdata['categoryid']] = str_repeat('&nbsp;', $catdata['depth']*2) . serendipity_specialchars($catdata['category_name']);
           }
           return $flat_cats;
+        } else {
```
garvinhicking: see above

include/functions_routing.inc.php
```diff
@@ -47,7 +47,7 @@ function locateHiddenVariables($_args) {
         } elseif ($v[0] == 'A') { /* Author */
             $url_author = substr($v, 1);
             if (is_numeric($url_author)) {
-                $serendipity['GET']['viewAuthor'] = $_GET['viewAuthor'] = (int)$url_author;
```
garvinhicking: I believe there was a reason to also have ?viewAuthor=XXX instead of only ?serendipity[viewAuthor] at some place. Did you check that also possible other plugins might reference this $_GET['viewAuthor'] variable?

plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
```diff
@@ -401,547 +401,547 @@ function clean_nl2brtags(&$entry)
...
+     * @param string text
+     * @return text
+     */
+    function tag_clean($textstring)
+    {
+        $text = str_split($textstring);
+        $tagstart = false;
+        $tagstart_b = false;
+        $tagdef = false;
+        $endtag = false;
```
garvinhicking: Please try to not mix code formatting and code logic commits; a commit like this is hard to review. Please first commit only the code formatting change and then possible logic changes to better distinguish this, if possible?

plugins/serendipity_plugin_categories/serendipity_plugin_categories.php
```diff
@@ -255,6 +255,9 @@ function generate_content(&$title)
                     }
                 }
 
+                // strip language tags if multilingual plugin is active
+                serendipity_plugin_api::hook_event('multilingual_strip_langs',$cat['category_name']);
```
garvinhicking: see above

plugins/serendipity_plugin_superuser/serendipity_plugin_superuser.php
```diff
@@ -46,7 +46,8 @@ function generate_content(&$title)
             $base = $serendipity['serendipityHTTPPath'];
         }
 
-        $link = $base . ($serendipity['rewrite'] == 'none' ? $serendipity['indexFile'] .'?/' : '') . PATH_ADMIN;
+        //$link = $base . ($serendipity['rewrite'] == 'none' ? $serendipity['indexFile'] .'?/' : '') . PATH_ADMIN;
+        $link = $base . 'serendipity_admin.php';
```
garvinhicking: Why did you change this? In rewrite environments it would be nice to use a short-hand established link like /admin?

include/functions_entries.inc.php
stephanbrunker: The functionality of the hook is to remove the translation tags. If there is a result array passed via the hook, is it safe to purge the tags from all the result fields? That was the question, because the issuing queries are different. Or I have to pull the values to be stripped out of the array in a new one, put that into the hook and reinsert the result. Seems doable.

include/functions_config.inc.php
stephanbrunker: I did a full-text search in the core, core-plugins and additional plugins and it returned empty.

include/functions_routing.inc.php
stephanbrunker: As far as I could determine, the RSS feed issues the requests outside of the [serendipity] array, i think this is because of the RSS interface, and because of that I left the serveFeed() function untouched. viewAuthor is not set directly into GET from any other function or plugin. I did my little purge here because some of the 'lang' variables were used in $serenditpity['GET'] and $_GET at the same time and written to which really created a mess. Ideally, all GET should be moved into a [serendipity] array, but there are usages in admin/plugins.inc.php, exit.php and comment.php where i couldn't identify the source of and because of that i didn't touch them.

There is also the occurrence of $_GET['p'] in serveEntry() which I don't know what that is for. That means you can get an entry with `/index.php/?p=33`   where p is the entry id. But basically, it would work even with `/index.php/?33` .

plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
stephanbrunker: Sorry for that. I did only the formatting and changed nothing else. I fixed that after I found the formatting rule (indent by spaces) in the code primer. Because I never got invited to work in a team before, I simply don't know about the workflow.

plugins/serendipity_plugin_superuser/serendipity_plugin_superuser.php
stephanbrunker: My debugger monitored the initialization of the serendipity framework and I found out that by clicking on the 'login' link, the call was first to `/index.php/?admin` only to get rerouted to `/serendipity_admin.php` which did all the initialization again which was also part of the lang setting problems. Because the code primer said that all the routing for the backend goes via the /serendipity_admin.php it seemed logically that the link should point there. Sorry if I did not get the concept of the rewrite engine, but in the end, the link should not point at index.php, according to the documentation.

th-h:
> It might be a bit dangerous though to put this inside the to-be-released s9y version, and we should have at least a test candidate for these changes for others to test. 

Of course. It's something for 2.4, I think. We can backport changes bit by bit to 2.3.x later on, after they got some testing.

include/functions_entries.inc.php
garvinhicking: Yes, you can pass $cat as $eventData to the hook. It's passed by reference, so any modification the plugins make then to $eventData on the hook will continue in the loop of functions_entries.inc.php to hold that modified data. Then only your plugin needs to make the foreach loop on $eventData, and the whole plugin API introspection is only called once instead of X times.

include/functions_config.inc.php
garvinhicking: Sometimes I am afraid of places where dynamic array access could be used. Like:

```
$whatToCheck = 'content';
if (condition()) {
    $whatToCheck = 'interface';
}
echo $serendipity[$whatToCheck . '_lang'];
```

I did learn that this is bad code, but it is still left in a couple of places, and impossible to search for. So I'm leaning toward the safer variant to leave the variable in there, it shouldn't hurt but shouldn't break things. Other opinions maybe? Of course the wish to only leave relevant code is important as well.

include/functions_routing.inc.php
garvinhicking: OK, understood. Well, then just lets try to remove that viewAuthor and see if it breaks things ;)

That "?p" variable is Wordpress compatibility simply, for people who moved from one system to the other by keeping entry IDs and without needing specific rewrite rules for compatibility.

plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
garvinhicking: All good, I don't wanna come off as picky, it's just that when I try to review code, I try to really see the changes. In an ideal world, we wouldn't mix multiple commits into one as well, but I see the overhead in splitting up this larger commit to multiple variants. 

To go on, please feel free to first make a merge request on just the formatting changes if possible. Then we merge that, and then we can merge the rest of the actual code changes without this overhead of formatting commits?

plugins/serendipity_plugin_superuser/serendipity_plugin_superuser.php
garvinhicking: OK in that case, yeah - for non-RewriteEngine rewrites it does mean a separate reroute. But when URL rewriting is enabled, it directly maps to the right URL. It wasn't a good design decision to prefix many files with "serendipity_", so where possible I'd like to keep the URLs more clean :D. 

So the best variant should be to use serendipity_admin.php when URL rewriting is off:
```
$link = $base . ($serendipity['rewrite'] == 'none' ? 'serendipity_admin.php' : PATH_ADMIN); ```

that could work?

include/functions_entries.inc.php
stephanbrunker: What I meant is that it is not so simple. $cat may be a multidimensional array and then I'd need to expand the tag removing function on the other side that it iterates through the whole array. But that may have unintended consequences because I only want to remove the tags from the 'category name' and 'category description' fields and from the whole query result. the tags may be quite unique, but who knows … And because the queries are not consistent, the fields can also be numbered ('assoc' option or not). To be safe, I have to copy the fields to strip out of the result array, pass them to the hook and reinsert them afterwards. A little more code, and because of that it was easier to iterate inside the source function and do the multiple calls to the hook. But as I said, I can do that (maybe this evening).

garvinhicking: You could do the distinction on the event hook itself? It knows the context and what is passed, so you could pass 'addData' along to specify to the plugin that it needs in which context the plugin hook was called. Another way is to not use one central hook, but use one hook for multi-dimensional arrays and one for the other?

include/functions_config.inc.php
stephanbrunker: I removed it and it didn't break anything. Because I started with the problems caused by inconsistent language settings, I would be glad if something shows up which depends on that variable because I need to fix that, too. I have so far condensed the language variables into these:
$serendipity['autolang'] : the name is misleading, this is the ultimate fallback language which remains 'en' in all cases. 
$serendipity['lang']: the language to be shown for every output
$serendipity['default_lang']: the default language of the blog
$serendipity['detected_lang'] the language determined by user input or browser negotiation, if that fails, NULL.
$serendipity['GET']['user_language'] the language requested for the blog by user
$serendipity['GET'['lang_selected'] the language requested for an single article by user

I'd prefer to clean it up as much as possible. If there is a call to that variable, this would be a very clear error message and a simple replace with one of the above.

plugins/serendipity_event_nl2br/serendipity_event_nl2br.php
stephanbrunker: It was just this one and next time, I won't mix, I promise ;-)

include/functions_config.inc.php
garvinhicking: OK, you convinced me. Let's remove it.

stephanbrunker:
I did a commit with the requested changes now. But there is one issue unsolved and I don't know yet how to tackle it: If a user has chosen a language in the frontend which is different from his default language, the language changes while changing to the backend. That is intended. The problem is the order of events in the _functions_config.inc.php_: First, _getSessionLanguage()_ looks for the language to display, based on user input. Then _serendipity_login_ is called which first calls _serendipity_authenticate_author_ and this the eventhook _backend_login_ . And this as the first eventhook initializes the plugins and at that time, the language for the user isn't yet determined - because the real login check is happening after that call. The user language setting is done after _serendipity_login_ returned with _serendipity_load_configuration_ and _serendipity_getPostAuthSessionLanguage_ . The effect is that the backend is loaded in the correct user language, but some output depending on plugins is in the frontend display language. This is only until the next reload, then the session variable switched by _getPostAuthSessionLanguage_ effects _getSessionLanguage_ .

In short: How can I make the plugins load the correct language for a authenticated user which is not yet logged in when the plugins are initalized?

garvinhicking:
I'll merge this now. If there are objections, we can of course refactor this commit.

@stephanbrunker I must admit I am not properly familiar with the code at this point to help in your remaining question. I'd probably rather leave it like the current situation, where some elements might show a wrong language after first login, when its fixed at the second page request. This should be safer than what it might possibly break with the other change you propose.

I am thinking if maybe a valid solution would be to detect if the initial language is different from the one after the proper login, and if that's the case, do a simple HTTP redirect to the backend page again, so that the user won't see this output. The downside of this could be that possible login errors or other actions that are performed after the initial login might not be shown on the second call to the backend, so I'm also not very sure if we would make things worse with that.

Issue #657, opened 2019-10-18
External plugin login


stephanbrunker:
There is one language bug left for me to fix. The language in the backend is now plucked from the user language setting in the database, in the frontend can the user still select any language. The bug is now when this language change happens, the plugins are loaded in the frontend language. This is because of the order of events in the configuration. First, the general language is determined with `serendipity_getSessionLanguage()` , then `serendipity_login(false)` is called and when a user is logged in with `serendipity_load_configuration(authorid)` the language for the user which is reverted to the original selection for the frontend in `serendipity_getPostAuthSessionLanguage()` . Then `lang.inc.php` is called which sets the language. 

The parameter 'false' for serendipity_login should not - following the description - forward the login to external plugins. There are three event hooks called: backend_login , backend_loginfail and backend_auth. The last one is really depending on the false parameter, the others not. The issue is only that by calling any event hook at this point in the configuration process, all the plugins are loaded and by that the language constants defined, which by nature cannot be changed later. And prior to the login, the user language is not yet determined. `serendipity_login(true)` is called again later on by serendipity_admin.php.  

I have checked now which plugins use these hooks and why. backend_loginfail is only used by event_externalauth for logging purposes. backend_login is used by event_browserid and event_openid. backend_auth doesn't matter because it is issued after all the language settings because of the parameter.

My proposal would be to expand the false parameter to the other two event hooks, that these external plugins are called after the framework is set up like intended. By doing that, the first event hook is frontend_configure at the end of the configuration which is a good place to initialize the plugins. 

For event_externalauth it shouldn't matter because I think the logging of failed attempts is intended for the own attempt by backend_auth which is not called on the first run. I am not sure if both other plugins should login at backend_login or at backend_auth like the externalauth plugin, but anyway, it should not matter when that also happens on the secondary login run. Basically, this long message revolves around two small if() conditions. And to load the languages correctly for users authenticated with an external plugin, after the second run of serendipity_login the page should be reloaded in case of an login, which is simply done by sending a header. Then the login session vars and cookies are set and the user language is loaded correctly prior to plugin initialization.

onli:
Hi @stephanbrunker! I think these hooks are not used that much. If you want to change details there the alpha cycle is the right moment, if it has unforeseen consequences we should find them. 

This change is not already in #658, right?

stephanbrunker:
[...]

PR #658, opened 2019-10-20
Spamblock plugin / last multilingual bugfix

stephanbrunker:
I have divided it into two commits: One is the fix for the spamblock plugin: It now uses a timeout between showing the page for reading and allowing contents. You can use a fixed amount of time or an adaptive one, depending on reading and writing speed. That is tested and works, now I don't have any spamcomments anymore, but just got a real one with the adaptive setting to 500 words/chars per minute. To circumvent that, the spammers need to keep open each connection for minutes and that probably slows down their operations a bit because they hopefully can't have that big a number of connections parallel. 

The second one is changing the names of the languages to their own languages as was requested in the forum. And fixing the last remaining language bug which was the initialition point of the plugins -> Issue #657 

That second one should be tested by someone who has openid plugin installed. I don't have and don't have openid able accounts, but it should work anyway.

### DIFF
--- a/include/functions.inc.php
+++ b/include/functions.inc.php
@@ -1060,6 +1060,7 @@ function serendipity_addCategory($name, $desc, $authorid, $icon, $parentid) {
         'category_description' => $desc
     );
 
+    serendipity_plugin_api::hook_event('multilingual_strip_langs',$data, array('category_name','category_description'));
     serendipity_insertPermalink($data, 'category');
     return $cid;
 }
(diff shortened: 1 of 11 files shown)


onli:
Ok, I'll merge this now. Please in the future try to make one dedicated PR for each change. For example the nl2br change does not belong to this. Probably you just forgot to switch back to the default/a different branch after making the PR?

th-h:
Yes, it's really a bit unfortunate to mix three unrelated changes in this way. :-/

I'll try (!) to document them accordingly in NEWS/Changelog (and hope I understand what has been done ...)

stephanbrunker:
Sorry, I have no experience with project workflow if you work in a team and make it up on the seat of my pants ;-)

Upside is: I can only learn from everything I do. In this case: creating several forks in my git and merging them only after the PR is merged upstream. For the first PR, i added everything to the changelog, I just forgot to do that with the second one. For the additional plugins, I added everything into the changelog files.

th-h:
No problem, and sorry for sounding a bit grumpy. :)

stephanbrunker:
Also no problem, I made changes in a lot of areas … that makes it complicated.

PR #662, opened 2019-11-15
functions_routing.php: set action to empty when serving JS

stephanbrunker:
I discovered that by working on the nl2br plugin: the event hook 'frontend_display' was called multiple times for every page shown. Also the content to be processed by the plugin was the complete list of recent entries.

The issue originated by a call to the index.php with a serendipityJS request. In include/routing.inc.pnp the serveJS function called genpage in a hacked fix to init smarty, but no action was given. This caused genpage to generate the default content which eats resources and generates additional loading time. I fixed the fix by defining the action to 'empty' which leads to no page generation while serving Javascript.


onli:
Looks like a good solution!

PR #663, opened 2019-11-15
nl2br: a-elements and html comments

stephanbrunker:
I added the functionality that a-elements can contain everything from inline elements to block elements (which is html5 standard)

The plugin now also ignores everything inside html comments.


onli:
This might be very nice, thanks

PR #665, opened 2019-11-20
Don't allow requesting an archive page that doesn't exist

lotharsm:
Currently, it's possible to request archive pages with a page number higher than the actual page count.

For example, look at https://blog.s9y.org/archives/P999.html - it's requesting page 999 from 47. Furthermore, if you request a *really* high page number, the SQL query will break.

This patch adds a simple check and overwrites the page number in ['GET']['page'] with the highest page number that's currently available.


onli:
Nice and elegant solution. Thank you!

PR #668, opened 2019-12-15
admin template entrylist: entry title was double escaped

stephanbrunker:
It was the first time I used html entities in a article title. There were no issues, but in the backend entry list, the title was double-escaped, the entities shown as &lt; etc. This is unnecessary and due to an |escape option in the template.

### DIFF
--- a/templates/2k11/admin/entries.inc.tpl
+++ b/templates/2k11/admin/entries.inc.tpl
@@ -142,7 +142,7 @@
                         </div>
                     {/if}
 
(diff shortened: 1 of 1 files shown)


onli:
Did you pinpont where the first escape is happening?

stephanbrunker:
I didn't follow the entire line of programming from the edit window to the entries display. You are right, it looks now that entities aren't escaped at all in the title. The link url is stripped, that is okay. In the frontend template (2k11) the $entry.title is also not escaped. I cannot find the processing of the entry title, but it seems like it is escaped because html code in the title doesn't work. I also know that my nl2br plugin does escape entities in the body, but this is not shown in the phpMyAdmin database entry, so it is difficult to say. But if the |escape option is not in the frontend, that means it shouldn't be there in the backend too?

onli:
Hey, sorry for the late reaction. You're right. This is double escaped and it would be more consistent if that were not. I went through the git history and the escape is there since the beginning of the smartified backend, it was not added in response to a security issue. So I will merge this now :) Thank you!

PR #676, opened 2020-03-17
changed parameters order for implode()

surrim:
Passing the glue after the pieces (i.e. not using the documented order of parameters) has been deprecated


onli:
Thanks!

PR #683, opened 2020-03-25
Update spamblock query to work in MySQL 5.7+ with default "strict" mode

fasterit:
The old query will fail with ONLY_FULL_GROUP_BY set which is default from
MySQL 5.7.5 onwards. This results in the .htaccess file not being updated.

Also adding some fencing to the foreach to prevent php error messages.

Cf. https://board.s9y.org/viewtopic.php?f=11&t=24753


onli:
I was only able to test this on mariadb 10.3.18, where both queries work, but both have basically the same effect. You add max(timestamp) to the result array, but the ip is taken via that for loop anyway, so that should be fine. Thanks for the PR!

onli:
PS: Next time please also up the version number :) I will do that now.

Issue #686, opened 2020-03-31
update bundled jQuery to 3.x


mmitch:
I've just found the Google Lighthouse plugin for Firefox and ran it against my blog which is already updated to Serendipity 2.3.4.
The Lighthouse report mentions 2 medium security issues in the jQuery 1.12.4 that is bundled with Serendipity 2.3.4: the two red lines at https://snyk.io/vuln/npm:jquery?lh=1.12.4

As far as I can tell:
* jQuery 1.12.4 is the newest version of the 1.x branch
* both jQuery 1.x and 2.x don't receive any patches any more
* there is a [jQuery migration](https://github.com/jquery/jquery-migrate/#readme) tool from version 1.12.x to 3.0

I don't know much about JavaScript or jQuery – is an update or version switch relatively painless or would this mandate a rewrite of half of Serendipity's frontend?

_(According to the contribution guidelines, I have contacted Garvin by email to check whether this is security relevant – his quick triage/response was: 1. the bugs in jQuery are more on the cosmetic side 2. updating jQuery might break plugins 3. I should go on to post this issue here on Github: [x] done :-)_

yellowled:
>  is an update or version switch relatively painless or would this mandate a rewrite of half of Serendipity's frontend?

I have to admit I haven't updated anything from s9y 1.x to newer versions in a long time. The issue is not so much s9y's frontend, but plugins (as Garvin mentions). Which, of course, is a matter of testing first and foremost.

* https://jquery.com/upgrade-guide/3.0/ lists all breaking changes
* We could of course bundle the migrate plugin for the time being in order to be able to upgrade quickly, but that does not spare us the testing.
* The 3.x migrate plugin clocks in at 8.8 KB minified, which is not that bad. However, it's an additional HTTP request.
* We would of course have to do a release anyway.

The most likely issue is probably third-party stuff breaking. Some outdated, unmaintained script that one of our plugins or themes uses. I'll try to gather information on which themes/plugins even use jQuery. (I don't think dropping jQuery altogether is an option.)

yellowled:
**Core themes that use jQuery**

* clean-blog
* next
* default (I think, I'm not 100% sure)
* 2k11, including in the admin interface
* bootstrap4 (but that *should* work with jQuery 3.x since bootstrap bundles that)
* timeline

[...]

PR #687, opened 2020-03-31
Migrate to jQuery 3.x

yellowled:
**Note:** This should be tested before merging! See referenced issue for a list of core themes and additional plugins depending on jQuery.

- Update bundled jQuery to 3.4.1
- Add jQuery migrate plugin
- Include jQuery migrate in all frontend and backend templates

References #686


onli:
@yellowled How can I test this for you? I assume this is still something that should be merged?

yellowled:
@onli Basically, see #686 and the comments there. I did some testing, as did @mmitch and @surrim, but there are (additional) plugins that will break, most notably [`serendipity_event_commentedit`](https://github.com/s9y/Serendipity/issues/686#issuecomment-623154929).

I think the only way to be sure would be to test *every* plugin and/or functionality that requires jQuery to see if it requires an old version and/or if using the migrate plugin helps. In some cases (like commentedit) this seemed to require some familiarity with the – PHP – code or at least the plugin that I did not have. That's why I gave up testing.

Also, the version should be bumped to jQuery 3.6.0.

I'm not sure "how much" functonality just merging would break at this point. Probably not too too much in core, core plugins and core themes, but that's just a guess at this point.

onli:
I gonna have a look at commentedit. It was probably too complicated for me at the time when I wrote it, but it's a feature I still want. Maybe I can clean it up and fix the bug at the same time.

Most of the other plugin errors linked in that thread seem to be PHP issues.

onli:
I think we can actually merge this now. This PR now contains the current jQuery version, plus the current migration plugin. With commentedit fixed I had a look at a few of the other plugins marked as problematic (livecomment, lightbox, linktrimmer), and they all worked - or had issues on the PHP level. PHP errors some of the bugs opened in #686. 

I didn't even notice output from the jquery migration plugin, but I might have just been unlucky with the plugin selection for that - and either the backend JS code is completely compatible, or I'm misunderstanding how that plugin is supposed to work.

In any case, I think merging this into master and testing it that way is the only path forward. Ok @garvinhicking ?

fe-hicking:
Hm, theres a .orig file mixed in the commit.

I also dont know what jquery-migrate is, I wonder if we could do without...?!

onli:
> Hm, theres a .orig file mixed in the commit.

I removed it.

> I also dont know what jquery-migrate is, I wonder if we could do without...?!

https://github.com/jquery/jquery-migrate/#readme explains its function (and also why I did not see console warnings, the production version does not output them...). It paves over the breaking changes. I think we need it for now. We could try to rework the core JS for the changes, but then there would still be all the plugins left to fix...

onli:
@garvinhicking Do we go ahead here as well? :)

onli:
Actually, we might be able to skip jquery migrate. I'll provide some commits if everything works out.

PR #689, opened 2020-04-06
[plugin recententries] Fix multilingual issues

stephanbrunker:
I had to implement the multilingual SQL in another way. Instead of using the fetchentries event hook, the multilingual data is now fetched directly from the entryproperties table if the multilingual plugin is installed. With the last version of the multilingual and customarchive plugin (additional_plugins/multilingualfix branch), you can switch now between displaying all entries (fallback to default language if untranslated) and display only the translated ones. This change doesn't affect anything outside the multilingual plugin.

### DIFF
--- a/plugins/serendipity_plugin_recententries/ChangeLog
+++ b/plugins/serendipity_plugin_recententries/ChangeLog
@@ -1,3 +1,6 @@
+2.7.1 * because of changes in the multilingual plugin, 
+        get the options out of the plugin and hard code the SQL for the 
+        multilingual data set
 2.7
     * using the sql condition set aquired by event_hook::'frontend_fetchentries'
       to get the multilingual data set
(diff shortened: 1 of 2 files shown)


onli:
@stephanbrunker Is this safe to be merged, or does it depend on a second change?

stephanbrunker:
Yes, it is safe to merge. The changes I made with 502b837 weren't compatible with the not yet published latest version of the multilingual plugin and the idea of displaying only the translated entries. So I changed that partially back to the previous way where the plugin simply fetches the multilingual data directly from the database instead of getting the multilingual SQL conditions indirectly via an fetchentries event hook. 

The only downside is more philosophically that the multilingual plugin is not really a simple independent plugin but has some of its functionality prepared by the core. Just do a text search of 'multilingual' in the /include folder. But it would be more or less impossible  to remove all the links to the multilingual plugin from the core and do everything with eventHooks so you can do it the far easier way and like in the commit above hardcode the multilingual display directly in these plugins. If multilingual is not installed, nothing happens.

th-h:
Not much time these days, so just a short remark:

> The only downside is more philosophically that the multilingual plugin is not really a simple independent plugin but has some of its functionality prepared by the core.

I share your opinion - it makes sense to have multilanguage support in core, as we do have support for e.g. Spartacus in core.

stephanbrunker:
Part of the situation is that the multilingual data rides piggyback on the entryproperties data which is more or less part of the core. fetchEntry() odr fetchEntries() get the properties en bloc and with them the multilingual translations. So the data is already fetched and the question is only what to display. In the case above: query the plugin config var and modify the query for the last entry titles. 

@th-h But it makes sense to hide all the multilingual options for the user which uses only his default language. So, activating them by installing the plugin makes sense. But avoid to make it unnecessary complicated by relying on eventHooks as I tried with the now reverted commit.

onli:
Great! :)

PR #690, opened 2020-04-06
[plugin entryproperties] removed cache from serendipity['entry'] arra…

stephanbrunker:
…y, is already in subarray

While debugging, i discovered that the entry data is three times in the 'entry' array: One time in [body] etc, one time in the properties subarray as ep_cache and a third time the same cache directly in the array. The last one is unnecessary because the fetch_entries function put it already there, no need to use the fetchentries event hook to put it there a second time.

### DIFF
--- a/plugins/serendipity_event_entryproperties/ChangeLog
+++ b/plugins/serendipity_event_entryproperties/ChangeLog
@@ -1,3 +1,9 @@
+1.41.5
+------
+    * removed cache SQL from fetchentries event hook, cache is already put in 
+      properties subarray by issuing calling function
+
+
 1.41.4:
 -------
     * removed 'WHERE' keyword from fetchentries event hook, because that should be handled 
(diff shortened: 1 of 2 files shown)


stephanbrunker:
While debugging another issue in the nl2br plugin, I discovered that the 'frontend_display' eventHook is called for every entry while displaying the index page and never for displaying an single article. This is connected with the entryproperties cache. For gaining faster access time, when activated, the cache should load all the entries from the cache instead of processing them through the plugin. In case of the article (serendipity_FetchEntry), it  works correctly, the ep_cache_body is inside the properties subarray and because of that, the plugin API doesn't call the plugin again. 

Unfortunately, for the entry list: serendipity_fetchEntries, this is buggy. the cache doesn't appear in the properties subarray. To make it similar with fetchEntry, i added a call to fetchEntryProperties but that doesn't resolve the problem because now the properties - with the strange exception of ep_access - disappear between the start of serendipity_printEntries and the foreach-loop of the then grouped entries. 

So, this PR needs additional work to debug and clean up the issue. Ideally it should clearly separated between direct calls to fetchEntryProperties, the entryproperties eventHook and the fetchEntry/fetchEntries eventHooks. 

And a side notice: the description for the second argument of fetchEntries is incorrect. It tells that with false, only the body is fetched. In reality, it then fetches only the title, and thus all the calls are with true which is unneccessary for the index page where only the bodies are displayed. Something to do … ideally, the entryproperties will only fetch the cached body if only the body is in the entry array and similar to that the multilingual should only fetch the translated body, too.

stephanbrunker:
After hours of a goose chase, i could catch the culprit: The 'entry_display' eventHook of the multilingual plugin kills the cache. That means I have to create a separate multilingual cache and clean that up, but now I have an idea what to do.

PR #697, opened 2020-04-21
Switch from Zend to Laminas

soatok:
Zend Framework is dead. Long live Laminas!

https://getlaminas.org


onli:
Hi. I think we used zend only for a db layer experiment, in https://github.com/s9y/Serendipity/commits/master/include/db/generic.inc.php. That's something we never exposed. So while the PR looks completely legit to me - thanks @soatok ! - what we might actually want to do here is to drop zend and the generic db layer file.

onli:
Let's merge this for new and maybe drop this later. Looking again at the db layer it's still a nice approach.

PR #700, opened 2020-04-24
simplify cookie option code and set security flags httponly and sames…

hannob:
…ite=lax

This is a proposal for some security improvements for the session cookie parameters.

First of all I think the existing code had some unnecessary complexities. It first reads out options via session_get_cookie_params(). However there is no other place in the code that sets these parameters, so there should be no reason whatsoever that there are already existing settings that need to be preserved.

The "? true : false" is basically a noop, if we want to set true/false based on a boolean condition we can directly set the condition.

Then the real changes: Set the httponly cookie flag, which is a protection against some forms of XSS exploits (it makes the cookie unavailable to javascript) and lately also a protection against certain forms of spectre-like speculative execution attacks (because the cookie is no longer held in the memory browser rendering process).

I hope there is no javascript code within s9y directly trying to access the cookie, but I haven't seen any obvious breakage so far. Though this should get some testing probably.

Then I'm setting the samesite=lax flag which is a new protection against CSRF attacks. It basically means dangerous cross-site requests (e.g. POST requests) will be sent without the cookie.

I guess the most controversial part of this change is that the array-setting and samesite are only supported with php 7.3 and above. I'm inclined to say "PHP 7.2 is EOL anyway", but... I guess that's controversial :-)

I wanted to provide this for discussion first, but if the PHP 7.3 issue is the only thing keeping this from getting merged I will rework it with some if's so it will still work on older php versions. (Of course that would mean no samesite cookies for older PHPs.)

### DIFF
--- a/serendipity_config.inc.php
+++ b/serendipity_config.inc.php
@@ -10,9 +10,13 @@ if (defined('S9Y_FRAMEWORK')) {
 if (!headers_sent() && php_sapi_name() !== 'cli') {
     // Only set the session name, if no session has yet been issued.
     if (session_id() == '') {
-        $cookieParams = session_get_cookie_params();
-        $cookieParams['secure'] = (isset($_SERVER['HTTPS']) && strtolower($_SERVER['HTTPS']) == 'on' ? true : false);
-        session_set_cookie_params($cookieParams['lifetime'], $cookieParams['path'], $cookieParams['domain'], $cookieParams['secure'], $cookieParams['httponly']);
+        $secure = (isset($_SERVER['HTTPS']) && strtolower($_SERVER['HTTPS']) == 'on');
+        if (PHP_VERSION_ID >= 70300) {
+            session_set_cookie_params(array("secure"=>$secure, "httponly"=>true, "samesite"=>"Lax"));
+        } else {
+            // Support for PHP before 7.3, can be removed at some point
+            session_set_cookie_params(0, '/', '', $secure, true);
+        }
         session_name('s9y_' . md5(dirname(__FILE__)));
         session_start();
     }


th-h:
> I guess the most controversial part of this change is that the array-setting and samesite are only supported with php 7.3 and above. I'm inclined to say "PHP 7.2 is EOL anyway", but... I guess that's controversial :-)

It is. :) We just changed the minimum requirement to 7.0.

Debian Stretch for example has PHP 7.0, was the stable release up to january, will get regular support unmtil june and will still be supported for two more years via LTS. CentOS 7 (which will get support until 2024) still has PHP 5.4, and the current CentOS (which will be supported until 2029) just has 7.2.

I think we should aim to be 7.0 compatible as long as that is feasible and then raise the requirements slowly.

(And I'm not qualified to comment on the rest of your assessment, but I think you'll be right. :))

onli:
> I hope there is no javascript code within s9y directly trying to access the cookie, but I haven't seen any obvious breakage so far. Though this should get some testing probably.

I think there is. The default directory when uploading media for example is stored there, see https://github.com/s9y/Serendipity/blob/master/templates/2k11/admin/serendipity_editor.js.tpl#L479.

Bu I think we could change `serendipity.SetCookie`, the javascript function, to use localStorage instead?

> I guess the most controversial part of this change is that the array-setting and samesite are only supported with php 7.3 and above.

Yes, that's too early. Like @th-h said, distros don't move that fast. Please guard that with a version toggle.

> (Of course that would mean no samesite cookies for older PHPs.)

Actually there seems to be a workaround: https://stackoverflow.com/a/46971326 :)

hannob:
> I think there is. The default directory when uploading media for example is stored there, see https://github.com/s9y/Serendipity/blob/master/templates/2k11/admin/serendipity_editor.js.tpl#L479.

To clarify: This looks to me like it's using a cookie, but not the session cookie. The change I propose is for now for the session cookie only. (But we could and probably should look at the security settings of the other cookies s9y is using.)

> > (Of course that would mean no samesite cookies for older PHPs.)
> 
> Actually there seems to be a workaround: https://stackoverflow.com/a/46971326 :)

That's for setcookie, but not for the implicitly set php session cookies. (There may be a workaround for that, too, but... I guess I'm okay with "we provide this extra security feature only for users of latest php").

hannob:
Updated the code with a compatibility-if-clause for older PHP versions.

onli:
Oh, right. Alright then :) I'm gonna merge this now, so that if there are issues with that we notice it early enough.

PR #701, opened 2020-04-24
Restrict block display of summary to trackbacks.

th-h:
summary elements in the entry body should get a triangle symbol displayed (as we do in the backend).


th-h:
This went ... wrong.

PR #706, opened 2020-04-26
Set hashtype = 2 when converting from MD5

erAck:
Otherwise we'd end up with an attempt to authenticate with SHA1 against
BCRYPT that never succeeds, see
https://board.s9y.org/viewtopic.php?t=24720


onli:
Thank you, clear bug.

PR #707, opened 2020-04-27
Disable ckeditor ACF to fix #705

### LINKED ISSUE #705: Multi-insert of images (generating a gallery) doesn't work with CKE
### Steps to reproduce:

* Activate WYSIWYg editor (tried with "Srendipity Standard").
* Start a new entry, insert two or more images simultaneously.

###  Expected result ...

... would be this code:

```
<ul class="s9y_gallery plainList">
<li class="s9y_gallery_item"><a class="serendipity_image_link" href="[...]"><!-- s9ymdb:10 --><img class="s9y_gallery_image" src="[...]" alt=""></a></li>
<li class="s9y_gallery_item"><a class="serendipity_image_link" href="[...]"><!-- s9ymdb:9 --><img class="s9y_gallery_image" src="[...]" alt=""></a></li>
</ul>
```

### Actual result ...

... is this code:

```
<ul>
<li><a class="serendipity_image_link" href="[...]"><!-- s9ymdb:10 --><img class="s9y_gallery_image" src="[...]" alt=""></a></li>
<li><a class="serendipity_image_link" href="[...]"><!-- s9ymdb:9 --><img class="s9y_gallery_image" src="[...]" alt=""></a></li>
</ul>
```

We miss the classes for `<ul>` and `<li>`.

Tested on 2.3.5 and current master.

onli:
And to prevent similar issues in the future.

By disabling the html filter ckeditr has integrated, we prevent it from filtering html we want in the editor, like the media gallery classes. In fact, I can't think of a single use case where this filtering might be useful to us.

This did fix #705 in my dev blog and the editor seemed to still work as expected, but an additional testrun would be nice.

### DIFF
--- a/htmlarea/ckeditor_s9y_config.js
+++ b/htmlarea/ckeditor_s9y_config.js
@@ -13,63 +13,7 @@
  */
 CKEDITOR.editorConfig = function( config ) {
 
-    /** SECTION: ACF
-        Advanced Content Filter works in two modes:
-            automatic - the filter is configured by editor features (like plugins, buttons, and commands) that are enabled with configuration options
-                        such as CKEDITOR.config.plugins, CKEDITOR.config.extraPlugins, and CKEDITOR.config.toolbar,
-            custom    - the filter is configured by the CKEDITOR.config.allowedContent option and only features that match this setting are activated.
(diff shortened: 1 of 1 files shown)


onli:
1 month merge :) It's not a big change, we can easily revert if need be.

PR #709, opened 2020-04-28
Add Gitlab as Spartacus mirror

onli:
I created a gitlab mirror of our github repos at https://gitlab.com/s9y_blog. This gives us the chance to have an additional spartacus mirror.

### DIFF
--- a/plugins/serendipity_event_spartacus/ChangeLog
+++ b/plugins/serendipity_event_spartacus/ChangeLog
@@ -1,3 +1,7 @@
+2.39
+-------
+    * Add new gitlab mirror for plugins and themes
+
 2.38.4
 -------
     * Change dashboard upgrade notification to simple link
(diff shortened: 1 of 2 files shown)


th-h:
I think there is no automatic mirroring between Github and Gitlab? Perhaps iht would be better than to restrict the mirror to additional_plugins (and additional_themes) and delete the other Gitlab repos so we don't get contributions from there?

And we'd need to add the s9y-Bot to the Spartacus repos (and should perhaps restrict commit rights to the bot so we don't get accidental commits there).

onli:
There is automatic mirroring :) I configured all the repos to pull all changes from Github. That seems to be working fine, see https://gitlab.com/s9y_blog/additional_plugins/activity - my author name sadly on gitlab, but those changes are not triggered by me.

I don't think we thus need the s9y-bot there, since it's a 1:1 copy that includes the changes it makes on github.

th-h:
Ah, fine!

PR #712, opened 2020-04-30
escape filter words for regexp

hannob:
I noticed that using unusual characters ('[') in a filter keyword resulted in a regular expression error in the PHP logs (and no filtering).
The issue is that the filter is using regexps, but not escaping the filter keywords. Patch fixes this.


onli:
Nice catch!

PR #733, opened 2020-05-29
Feature subscription

stephanbrunker:
The subscription feature is now running as an alpha version on my live blog [www.intjblog.de](http://www.intjblog.de) and it looks very well. There are two version steps to alpha3 and alpha4 involved, the first was the subscription feature and the second one leading sections on the category and author pages. All the changes are in the NEWS file. So far, two things are untested: what happens to the links with the rewriteEngine on and the database changes for Postgres and SQLite. 

On the testblog, I already have subscribers and a few new articles were sucessfully emailed to them.


stephanbrunker:
Unfortunately, the merge is no fun. Two main problems:

- the /lang folder can be completely taken from my branch
- the /bundled-libs folder can also completely taken from my branch, because I did the same switch from zend to laminas in the composer.json file, but just yesterday and in the meantime a lot of files changed. 
- otherwise there shouldn't be many conflicts.

include/functions_entries.inc.php
```diff
@@ -112,13 +114,15 @@ function serendipity_fetchCategoryInfo($categoryid, $categoryname = '') {
...
+                         c.xmlimage,
+                         c.mailimage,
                          c.parentid,
                          c.hide_sub
                     FROM {$serendipity['dbPrefix']}category AS c
                    WHERE categoryid = " . (int)$categoryid;
 
-        $ret =& serendipity_db_query($query);
-        return $ret[0];
+        $ret =& serendipity_db_query($query, true, 'assoc');
```
onli, line 124: Why that change?

include/functions.inc.php
```diff
@@ -1119,14 +1121,16 @@ function serendipity_addCategory($name, $desc, $authorid, $icon, $parentid) {
  * @param   int     The new category subcat hiding
  * @return null
  */
-function serendipity_updateCategory($cid, $name, $desc, $authorid, $icon, $parentid, $sort_order = 0, $hide_sub = 0, $admin_category = '') {
+function serendipity_updateCategory($cid, $name, $desc, $authorid, $parentid, $icon, $xmlimage, $mailimage, $sort_order = 0, $hide_sub = 0, $admin_category = '') {
```
onli, line 1124: Honestly unsure about this. xmlimage looks to me like something that belongs into the sidebar plugin?

include/db/generic.inc.php
```diff
@@ -250,7 +250,7 @@ function serendipity_db_schema_import($query) {
             static $search  = array('{AUTOINCREMENT}', '{PRIMARY}',
                 '{UNSIGNED}', '{FULLTEXT}', '{FULLTEXT_MYSQL}', '{BOOLEAN}', '{TEXT}');
             static $replace = array('int(11) not null auto_increment', 'primary key',
-                'unsigned'  , 'FULLTEXT', 'FULLTEXT', 'enum (\'true\', \'false\') NOT NULL default \'true\'', 'LONGTEXT');
+                'unsigned'  , 'FULLTEXT', 'FULLTEXT', 'enum (\'true\', \'false\')', 'LONGTEXT');
```
onli, line 253: I don't think we can do this. Plugins and parts of the core this commit missed might rely both on the NOT NULL and on the default value.

I assume you made that change because the `subscribe` table should have default false? In that case I propose to add a `{BOOLEAN_FALSE}` to the schema instead of changing what's there.

include/admin/images.inc.php
```diff
@@ -28,6 +28,18 @@
     serendipity_restoreVar($serendipity['COOKIE']['serendipity_toggle_dir'], $serendipity['GET']['toggle_dir']);
 }
 
+$messages = array();
```
onli, line 31: That commit is already in master and should not be here.

plugins/serendipity_event_nl2br/Changelog
```diff
@@ -1,3 +1,4 @@
+2.21.7: * add some newlines in output to make it better humanly readable
```
onli, line 1: This commit is already in master and should not be here.

include/admin/entries.inc.php
```diff
@@ -141,6 +141,8 @@
                     $entry['categories'][] = serendipity_fetchCategoryInfo($catid);
                 }
 
+                serendipity_plugin_api::hook_event('multilingual_strip_langs', $entry['categories'], 'category_name');
```
onli, line 144: Unsure whether that's in master, but seems unrelated to subscriptions

include/functions.inc.php
```diff
@@ -716,7 +717,7 @@ function xhtml_cleanup($html) {
 function serendipity_fetchAuthor($author) {
     global $serendipity;
 
-    return serendipity_db_query("SELECT * FROM {$serendipity['dbPrefix']}authors WHERE " . (is_numeric($author) ? "authorid={$author};" : "username='" . serendipity_db_escape_string($author) . "';"));
+    return serendipity_db_query("SELECT * FROM {$serendipity['dbPrefix']}authors WHERE " . (is_numeric($author) ? "authorid={$author};" : "username='" . serendipity_db_escape_string($author) . "';"), true, 'assoc');
```
onli, line 720: Assuming the function was not broken before that looks like an API change we can't simply make. It might break plugins.

lang/langsorter.php
```diff
@@ -164,7 +165,7 @@ function synclang($dir, $lancode) {
 echo "OK\r\n\r\n";
 
 // process single file
-if ($lancode != 'all') {
+if ($lancode != '-all') {
```
onli: Seems unrelated to the subscription feature to me, might already be in master.

plugins/serendipity_event_nl2br/Changelog
```diff
@@ -1,3 +1,4 @@
+2.21.6: * add missing 'u' inline element
```
onli, line 2: Unrelated to subscription feature.

include/plugin_api.inc.php
```diff
@@ -172,6 +173,7 @@ static function register_default_plugins()
         serendipity_plugin_api::create_plugin_instance('@serendipity_plugin_archives');
         serendipity_plugin_api::create_plugin_instance('@serendipity_plugin_categories');
         serendipity_plugin_api::create_plugin_instance('@serendipity_plugin_syndication');
+        serendipity_plugin_api::create_plugin_instance('@serendipity_plugin_subscribe');
```
onli, line 176: I'd be more comfortable with not making it a default plugin directly and having it one cycle as opt-in.

stephanbrunker:
Sorry, to keep it clear: 
commits
0409717586e0053c7f415ebb0f1201a934b0bbab
9f0af1f2753655032a82b2517cdcebb4e8244236
0cf227f3352c999786de7daa055e53c9d759dca6
695055f50c3a7689d0b6ceb872ff3c6d628e5f02
8e24108b889f11bce454b0b33e1ce1f587145a42
e893914c7c2885505a9e22a13c3743bab2f6191b

are already in master via cherry-pick. So, I would propose to make a rebase of the feature_subscription branch first, that should take care of these cherry-picked commits and then most of the conflicts should be gone.

include/functions_comments.inc.php
```diff
@@ -1042,26 +1001,28 @@ function serendipity_saveComment($id, $commentInfo, $type = 'NORMAL', $source =
  * @param   string  The body of the comment that has been made
  * @return null
  */
-function serendipity_mailSubscribers($entry_id, $poster, $posterMail, $title, $fromEmail = 'none@example.com', $cid = null, $body = null) {
```
onli, line 1045: We can't simply rename that function. Plugins might use it.

include/db/generic.inc.php
stephanbrunker: We can do that safely. Your objection is valid, I didn't think of the additional plugins, where it is used in event_photoblog. I changed it there accordingly to `{BOOLEAN} NOT NULL DEFAULT 'false' ` if serendipity core < 2.4. No other occurrence of this element in the complete code basis. For the comments table, I did that already in db.sql with this commit.

onli: When core > 2.4, right?

The thing is: We don't know all plugins. While yes, probably everything is on spartacus, but there are some custom installations and plugins out there. Changing the core like that means we need to announce it, give time, bump the serendipity version to 3.x with the next release. That's okay if it has a big positive impact, but here I don't see the need.

If we just introduce {BOOLEAN_FALSE} we can have the subscription feature without all of this.

stephanbrunker:
In my opinion, investing time into improving the core only makes sense if you can actually improve it. That includes changing functions. I worked with enough software and it happened often enough that after a new release a lot of the external plugins quit working until their developers implemented the necessary changes. So, I am not going to add another case to a constant which has exactly four uses in all the code which is available to me when the logical solution is the one I used, because restricting a general 'boolean' to 'not null default true' simply doesn't make sense. The same is more or less true for the other changes to functions I made: It should have been so in the first place. 

I would even go so far and change the approach to the additional plugins completely. For example, for the last days I got every day an email from my Joomla system that I should update the core, and also get similiar mails that I should update the plugins because otherwise their functionality is not guaranteed. I made the changes over all the plugins in the core package, also for the additional plugins, and everyone else who is capable of writing their personal plugins should be very much capable of reacting to a changed order for the variable input for example if it is properly documented (perhaps I can do that explicitly). And for everyone who is trying to write their own plugins, how cryptic is it that {BOOLEAN} means 'boolean not null default true' and there is perhaps another one for 'boolean not null default false' and if he needs boolean null default whatever, then he needs to add another one or hack it ... nope. 

So, I would propose to bundle all the plugins for 2.3.x into one folder of the repository and create another folder 2.4.x with the plugins purged from all the old stuff and ideally only these where their developers tested them with the 2.4 core. The 2.4.x Spartacus then simply fetches the plugins from the 2.4 folder. That is the way it is handled almost everywhere I know, bundled by releases. Universal, release-independent plugins are the total exception and you simply don't know which ones are abandoned and which aren't.

Im truly sorry that I have to state that so plain, but I have invested a LOT of time since February on this project and intend to do my share of follow-up work. For my personal blog, I am already almost where I want because it has all the changes. But as stated in the beginning, my motivation depends on the possibility that I can shape it into the solution I envision (which is, by the way, something i mentioned on my blog because that is a result of my personality and enneagram type). And with that vision crumbling, the motivation crumbles, too. 

But I am also the type where the end result counts. To get there and make it easier for you, I researched all the git rebasing options and I can pull all these commits apart to order them strictly into feature branches, as long as nobody else worked with them. Just add the one library (because I don't know what happens if I run the composer update) and then I will do my part. 

And by the way: I truly think that s9y is "the best blog around", but I am afraid that its slice of the market is much thinner than the big ones like Blogger or Wordpress. The only major feature what I think is missing in the core are fixed assigned pictures for entries (normal and small version), that is something you see very often in themes for the ones above. Which would be easy to implement, too. But I think at least part of that is caused by the not-so-shiny front door: Someone who wants to start a blog would certainly look at the designs available and while the other two - or Joomla which is too complicated for a simple blog - have hundreds to thousands of templates, the number for s9y made in the last ten years is quite … small. Do you have any data about market share and actively attended blogs with s9y? While there are 38 languages total, the only fully translated ones are german and english …

And by the way … I just saw that I did implement a somewhat much simplefied version of the userprofiles plugin with just description and picture. But it doesn't cause a collision because if there is no content, my version hides automatically and doesn't interfere. But I needed a place outside of the authors and categories sidebar to put the subscribe and feed buttons, now you can hide them on the sidebars and only display them on the category and author pages which looks much nicer in my opinon. But you can simply configure that how you like.

onli:
Hi Stephan 
Let's chat in slack tomorrow or soon, that will be more productive. Please understand so far that my review is not meant to block the feature (not that I could do that if I wanted to), but to help it get in the core.

onli:
Hi @stephanbrunker. So I now finally had time to test this. It does not work in my dev blog, I think because the upgrader(?) relies on `ALTER TABLE`.  That's supported in sqlite in a very limited way. The result is a blog that can't display entries., probably because the upgrader subsequently failed to create the serendipity_subscriptions  table.

There is one `ALTER TABLE` in the upgrader function, but serendipity also tries to apply the **db_update_2.4-alpha2_2.4-alpha3_mysql.sql**. I have never used those sql migration files, I don't know whether it's a bug or a misconfiguration.

The upgrader also throws an error on the `CREATE TABLE serendipity_subscriptions` command. That's because of the  `timestamp int(10) unsigned DEFAULT NULL` column definition, sqlite does not like the unsigned. Well, it's in the mysql schema anyway.

---

I still think this feature should be implemented with less significant changes to the cores, this upgrader issues is a good example of the kind of issues we'd run into  with this otherwise. But I'd be happy to hear from @th-h and @garvinhicking what they think, specifically about the review comments made above, https://github.com/s9y/Serendipity/pull/733#pullrequestreview-423561198.

stephanbrunker:
That looks to be a specific problem only related to the SQLite version of the database upgrader. Because I really don't have any experience with SQLite, I just tried to hammer something together similar to existing stuff (hence the sql upgrade files) without the ability to test. So perhaps there are others with more knowledge in that environment able to fix these syntax problems and are of course better qualified to make the changes? The goal is well defined, it simply should do the same as the MySQL version which works fine as intended. (Btw, the same is true for the PostgreSQL version, which I also couldn't test). 

Because I don't have either an SQLite environment nor a database with actual data, I have to pass this one. Then it just condenses down into the question if the feature is worth having others do their contribution (which is marginal compared to my effort) or not, I did what I could do. Same is true for the changes in the core. I did what I think is the clearest and most logical approach, you are free to adapt it to your vision. I have put already so much time into it, it works on my blog (which was the original goal) and I don't intend to change it back to something which I don't think as logical and straightforward. 

Syntax conventions like the _if_ without brackets are another matter, there I did adapt. But I had a reason for what I did because I think that sometimes it is far more manageable in the long run to re-define a function than to stack a lot of exceptions on an existing one which becomes more and more cryptic as time goes by (the boolean thing is such a matter: instead of one case and several exceptions, it is much better to make a new, general one and define the special cases on demand). If you do that consequently from the start, you can crank down the complexity of a project a lot.

stephanbrunker:
One example of that general problem is for example that the SQL files and the SQL changes in the updater files are co-existing. Totally confusing. But the original intent - which is a good one - is that the SQL files are automatically and clearly separated by the different platforms, so everything which has different syntaxes should be done in the SQL files, even if it is possible to do in an updater script which additional checks. If there is only one version, the mysql.sql is executed (I did figure that out). Here, you just have to fix the *_sqlite.sql file and you are done.

onli:
@stephanbrunker I took from above that you worked a lot on this and at that point hadn't been interested to finish it. Maybe that changed till now? :) 

I read again through this PR now. The open comments I made above will continue to block this PR - like the SQL pattern change, no one will come fix them. At some point this would be closed.

Side-note: It looks right now that PHP 8 will force us to make a lot of code changes. A feature like this would profit a lot from being in master, so that we can fix it together. Keeping it as a separate branch will likely result in more work for you.

---

The feature is still a requested feature, but it should be possible to implement this with less breakage. Specifically, if you can transform this into a PR that:

 1. Add less lines to bundled_libs. 100K lines is unreasonable, where is all of that coming from anyway?
 1. Does not remove `serendipity_mailSubscribers`
 2. Keeps the existing SQL pattern replacements as-is
 3. Makes all of it a non-default config option, without setting a new default plugin

Also, process-wise, please:

 5. Answer each not outdated review comment
 6. Explain why we can not use the old subscribe column for this?

Then I would have another look at sqlite compatibility. At the very least I can propose a method that works with all db engines, maybe I could submit a fix for sqlite specifically.

stephanbrunker:
It is difficult as I have to work 7 days a week, 12-14 hours a day to keep my company afloat. So atm I don't have any spare time. 

- bundled-libs: I just added voku/html2text and made an update run because I don't know how to add just that single lib with its dependencies.
- all the features are non-default afaik
- For the author info and subscription feature I could wedge it on top of the existing page, same with the category page (optional and non-active by default)

But basically, the problem is that my style of work is not compatible with the s9y project as I have realized now as I do the crafting part of my buisiness. If I do something, I make it as clean and ordered as I can and I just cannot work otherwise. In the case of s9y, that meant that I needed to tidy up some scrambled and illogical parts of the core. My ambitions were not just the subscription feature, I also pulled the multilingual ability fully into the core - because there were already parts of it there and it just doesn't made sense to have it splitted in two locations. I did also a lot of work on the multilingual feature which was a long time stuck as a fork in additional_plugins until I did the merging into the core. Tidied also meant that I had to merge the existing comment subscription with the new blog, author and category subscription feature into one and I rewrote also the neccessary routing for it and reordered the database to make it as simple as possible. It would be a really pain to not touch the existing comment subscription and wedge the new features into the gaps. The result would be even more of a mess and a jumble of functions, spread over a multitude of database tables. I cannot work that way because you lose any resemblence of order.

To make sending an html formatted email possible, I had to rewrite the sendMail function too. 

After that, i realized that all mails sent have a multilingual problem: they were sent in the language active at the time of sending, which in a multilingual environment is not the language the subscriber has subscribed to. I started to fix that by splitting the language files into a static part and a dynamic one containing the strings needed for mailing. As I worked on the language files anyway, I cleaned them up too. By that time, the first PR was stuck and I lost interest to finish that part. Same was true with the other points on my ToDo-list: make the cache working with the multilingual plugin (the cache is simply killed atm if multilingual is active), tidy the cache (as it is doubled in the serendipity array in the root part and the entryproperties part), and last: adding the feature to link an image to an entry like in some other blog systems. 

Essentially, if I do add something, I'll go the extra mile and make it in a way like it was intended from the beginning and not attached later. For that, I fixed all the references for every single function and structure I changed, in the core and in additional_plugins (with version check). If I stumble over an issue or something nasty while doing that, I fixed that too. That is how I experienced big software products: they simply change what they seem neccessary and as a user, you have to adapt or quit. And oftentimes, they don't even fix bugs until the next (paid) version if something went wrong. I have a lot of examples for that from different companies. 

I have bigger problems now, so I made peace with that piece of work. You can accept my style of work (of course I will fix every bug that will show itself, except for non-mysql databases). In that case, I'll even add the other issues from my ToDo as soon as I can. Or you can drop the PR. I can live with both. As it is, my personal blog runs with a version with most of these features added: https://github.com/stephanbrunker/Serendipity/commit/bdf6b027949e0e8136aca5109855ae6c96ac83d0

But I fear that the two forks have already diverged too much to be merged. And I stacked the new features on top of each other because I didn't expect the first one to be stuck while working several steps ahead. And I cannot un-stack and cripple the already written code, that seems to be a waste of time which I don't have as the latest commit just runs fine.

stephanbrunker:
Just to add an example why I work that way, which happened more than once: My colleagues mounted some hanging cupboads. Some time later, they realized that they forgot to add the cables for the lights. They decided to make it the fast way and wedged them behind the cupboard without pulling them down. Even more time later, it was my turn and several issues could occur: Either I wanted to pull up the loose - which didn't work because the cable was stuck. Or the cable was shortened. Either way, I had to do what they didn't: disassemble the whole set of cupboards - which was a lot more work because of the additional steps made in the meantime -, correct it to the right way and assemble the whole lot again. 

That meant that I had twice the trouble and it took thrice the time compared to do it right on the first try. Related to s9y that means in my opinion: if I have to touch something which could have made better from the beginning, I fix that. Because it may happen that you have to touch that part sometime in the future. If you make patchwork, you only make it even harder for your future self to work with the result because over time, it increasingly messes up. 

So, if the database tables are more logical with all the subscription data combined, I'll do that. If there is a function which has the logical name for what I want to do and does part of what I want, I rewrite that function. If the arguments are impractical in the order before with a mess of code to circumvent, I'll bring them into the logical order, even if I have to track all the prior usages. 

The only exception to that style of work is when something has hundreds of usages and simply cannot be fixed without rewriting major parts of the code, I have to accept that. But I am not happy with it. But that is me, and it is fully in your right to have a different opinion, which just means that I cannot contribute.

onli:
I'll close here for cleaning up the PR list a bit, we sadly could never solve the conflict between backwards compatibility and the wanted code quality.

PR #740, opened 2020-08-14
Feature dateselector

onli:
This PR splits the datetime selector. 

Support to get a proper datetime-local selector in the browser is not happening outside of chromium, we have waited for years. But a date and a time selector is available in both chromium and firefox. By using those we can make it easier for users of Firefox without worsening the situation for Chromium users too much.

Alternative would be a polyfill, but I'm not aware of one that actually works. I also did not find a proper alternative datetime selector widget we could use instead (but there might be something available).

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,3 +1,4 @@
+   * Split date and time input in editor into two input fields
    * Improve performance of the media library by caching the file list
 
 Version 2.4-alpha2 ()
--- a/include/admin/entries.inc.php
+++ b/include/admin/entries.inc.php
@@ -60,17 +60,22 @@ switch($serendipity['GET']['adminAction']) {
         }
 
         // Check if the user changed the timestamp.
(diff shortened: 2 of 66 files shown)


yellowled:
@onli I currently don't have a stage to test this, but I don't really think there is a need to. I would also not "polyfill" the date widget since I'm not aware of a lean alternative. The JS looks good to me, the PHP I cannot review. Maybe have someone with a stage test this who can also review the PHP? Otherwise I'd say go ahead and merge.

onli:
@yellowled I sent you a pm on slack to my testinstance.

The thing more than functionality is that `datetime-local` has a nice widget on chromium and this solution is a step back for users of that browser. So what I'm looking for is a +1 on that change despite that, that we hit the right balance here :)

yellowled:
@onli I actually like this better because it clearly separates date and time. :)

PR #748, opened 2021-01-15
Unified parameter for "backend_image_add" hook

surrim:
* unified parameter for backend_image_add hook

### DIFF
--- a/include/admin/images.inc.php
+++ b/include/admin/images.inc.php
@@ -322,12 +322,12 @@ switch ($serendipity['GET']['adminAction']) {
 
                             // Insert into database
                             $image_id = serendipity_insertImageInDatabase($tfile, $serendipity['POST']['target_directory'][$tindex], $authorid, null, $realname);
-                            serendipity_plugin_api::hook_event('backend_image_add', $target, array('image_id' => $image_id));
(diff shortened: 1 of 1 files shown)


onli:
Hi @surrim, can you write a bit more about this change? To me it looks like a change that ought to be correct in theory but might have issues in practice, like there might be a plugin that only works with the $created_thumbnail string as the second parameter for the event hook. Does the change i nthis PR fix a bug somewhere for us?

surrim:
I grepped all plugins for the usage of this hook. It should be safe to change the parameter value because it's not used by any plugin. If it would be used by anything, the misbehavior would be reported because it's the same hook with different parameters.

I used it for updating the file size after compressing \*.gpx files (event_static_osm). Without the media id it's hardly possible to do that.

onli:
Grepping all plugins usually is not enough because plugins might very well not be upstreamed to us.

But this mismatch is too ugly and a possible source of bugs. Let's merge this now and keep it in mind if we get bugreports that might be caused by this.

Could you add a note to docs/NEWS describing the change?

onli:
Thanks again @surrim, as we talked about I documented this PR this time :)

PR #750, opened 2021-03-06
PHP 8 support

onli:
Enable Serendipity to work with PHP 8.

Incomplete, just a draft for now to make it better trackable. Smarty does not officially work with PHP 8 yet, but in practice seems fine. I updated to the newest version to minimize issues and applied a small patch that made HTML rendering work (in c0949c99345c88085d65025cabff2b91b2a6ab62).

This PR will contain a lot of changes in the core, but also in plugins and themes. The main issue is that we often access uninitialized variables and array fields, which in PHP 8 throws a warning. I don't think hiding warnings is feasible nor a good idea (in fact, the warnings made me correct a few real bugs that so far went undetected), so this PR is the result of combing through backend, frontend and default plugins to correct all code that throws warnings in PHP 8.


include/functions_config.inc.php
```diff
@@ -559,6 +559,7 @@ function serendipity_authenticate_author($username = '', $password = '', $is_has
 
         $rows =& serendipity_db_query($query, false, 'assoc');
         if (is_array($rows)) {
+            $is_valid_user = false;
```
garvinhicking, line 562: I believe thats a dual assignment, the lower one can be removed (2 lines down)

include/plugin_api.inc.php
```diff
@@ -1308,9 +1308,9 @@ static function load_language($path) {
         $probelang = $path . '/' . $serendipity['charset'] . 'lang_' . $serendipity['lang'] . '.inc.php';
         if (file_exists($probelang)) {
             include $probelang;
+        } else {
```
garvinhicking, line 1311: I think this can change the whole language flow quite a lot, if english language is not overloaded over a selected one?! IMO it should always fall back to english, because a translation might not contain new constants that have been added in the english file but not in the translation

serendipity_config.inc.php
```diff
@@ -210,9 +210,6 @@
 @define('USERLEVEL_CHIEF', 1);
 @define('USERLEVEL_EDITOR', 0);
 
-@define('VIEWMODE_THREADED', 'threaded');
```
garvinhicking, line 213: Where did these go?

include/functions_entries_admin.inc.php
```diff
@@ -104,10 +104,14 @@ function serendipity_printEntryForm($targetURL, $hiddens = array(), $entry = arr
 
             $template_vars['category_options'][] = $cat;
         }
+    } else {
+        $template_vars['category_options'] = [];
     }
 
     if (!empty($serendipity['GET']['title'])) {
```
garvinhicking, line 111: As I read this, this looks a bit XSS-prone to me, might need to check this?!

templates/2k11/admin/entries.tpl
```diff
@@ -42,11 +42,11 @@
             <button class="wrap_insimg" type="button" name="insJ" data-tarea="serendipity[body]">img</button>
             <button class="wrap_insmedia" type="button" name="insImage" data-tarea="serendipity[body]">{$CONST.MEDIA}</button>
             <button class="wrap_insurl" type="button" name="insURL" data-tarea="serendipity[body]">URL</button>
-            {serendipity_hookPlugin hook="backend_entry_toolbar_body" data=$entry_data.entry hookAll="true"}
```
garvinhicking, line 45: That looks like a major bug to me, doesn't this mean the event hook didn't work at all before?

include/admin/entries.inc.php
```diff
@@ -340,8 +340,10 @@
             foreach ((array)$serendipity['GET']['sort'] as $k => $v) {
                 $qString .= '&amp;serendipity[sort]['. $k .']='. $v;
             }
-            foreach ((array)$serendipity['GET']['filter'] as $k => $v) {
-                $qString .= '&amp;serendipity[filter]['. $k .']='. $v;
+            if (isset($serendipity['GET']['filter'])) {
+                foreach ((array)$serendipity['GET']['filter'] as $k => $v) {
+                    $qString .= '&amp;serendipity[filter]['. $k .']='. $v;
```
garvinhicking, line 345: We need to check XSS injection here too, that looks a bit dangerous?

include/admin/images.inc.php
```diff
@@ -20,13 +20,6 @@
 if (!is_array($serendipity['POST']) && $serendipity['GET']['adminAction'] == 'multicheck') {
     unset($serendipity['GET']['adminAction']);
 }
-// Listens on toggle_dir STRICT to list items per directory, or include all sub directory items
```
garvinhicking, line 23: This is still evaluated in the media_pane.tpl file?!

include/admin/images.inc.php
```diff
@@ -322,12 +342,12 @@
 
                             // Insert into database
                             $image_id = serendipity_insertImageInDatabase($tfile, $serendipity['POST']['target_directory'][$tindex], $authorid, null, $realname);
+                            serendipity_plugin_api::hook_event('backend_image_add', $target, array('image_id' => $image_id));
```
garvinhicking, line 339: That completely changes the API event input parameters, why?

include/admin/images.inc.php
```diff
@@ -407,12 +427,12 @@
 
                         // Insert into database
                         $image_id = serendipity_insertImageInDatabase($tfile, $serendipity['POST']['target_directory'][$idx], $authorid, null, $realname);
+                        serendipity_plugin_api::hook_event('backend_image_add', $target, $created_thumbnail);
                         $new_media[] = array(
                             'image_id'          => $image_id,
                             'target'            => $target,
                             'created_thumbnail' => $created_thumbnail
                         );
-                        serendipity_plugin_api::hook_event('backend_image_add', $target, $new_media);
```
garvinhicking, line 415: (see above)

templates/2k11/admin/plugin_config.tpl
```diff
@@ -18,7 +18,7 @@
 
             <div id="el{$config_groupkeys@iteration}" class="config_optiongroup{if $config_groupkeys@last} config_optiongroup_last{/if} additional_info">
             {foreach $config_groupkeys AS $config_groupkey}
-                {if $plugin_option[$config_groupkey]['ctype'] == 'seperator'}
```
garvinhicking, line 21: Thats more like a bugfix? Maybe better single those out as separate commits

plugins/serendipity_event_spartacus/serendipity_event_spartacus.php
```diff
@@ -701,7 +701,6 @@ function &fetchOnline($type, $no_cache = false)
             }
 
             xml_parser_set_option($p, XML_OPTION_CASE_FOLDING, 0);
-            @xml_parser_set_option($this->parser, XML_OPTION_TARGET_ENCODING, LANG_CHARSET);
```
garvinhicking, line 703: Why was this removed?

include/functions_routing.inc.php
```diff
@@ -428,66 +430,39 @@ function serveArchives() {
         default:
             $gday = 1;
 
-            if ($week) {
```
garvinhicking, line 431: I don't understand this fully yet, but doesn't it break a feature or misses some logic check now?!

include/functions_entries.inc.php
```diff
@@ -481,6 +481,7 @@ function &serendipity_fetchEntries($range = null, $full = true, $limit = '', $fe
     $ret =& serendipity_db_query($query, $fetch_single, 'assoc');
 
     if (is_string($ret)) {
```
garvinhicking, line 483: Probably accidentall, else it would also be information disclosure, we should only show that in debug envs

garvinhicking:
Great effort and work, that must have taken a lot of time. I added a few comments here and there.

I would prefer to not mix code cleanup commits with structural or logical changes (or bugfixes), those should get their own commits (and bugfixes getting their changelog entries; in one case the plugin API i.e. changed, that needs documentation as a breaking change). Else, if problems happen, it will be hard to understand why something changed if the commit message just looks like if it only touches on PHP8 compatibility.

As for adding isset() etc. for smarty, I am thinking if there's a way to broadly catch that on the Smarty engine layer, so whenever a variable is used, Smarty would do the check for its existance. Technically that should be possible, I'm not deep enough into Smarty these days to know, though.

include/plugin_api.inc.php
onli: That's impossible now. The moment a constant is already defined redefining it will throw a warning,  @ or no @. We have to change how we did that.

But you are right, this is a breaking change.

include/functions_config.inc.php
onli: No, this is needed. `if ($is_valid_user) continue;` will throw an error if the variable is not assigned yet.

serendipity_config.inc.php
onli: With the wind ;)

We run into a problem here as the constants were defined both as code for internal logic and in the language constant files for the layout. I first removed the internal constants, then re-added them when seeing what was up, but changed how the logic works as well. This has to be checked again with comments display in the frontend at a later time, and possibly be corrected.

include/functions_entries_admin.inc.php
onli: Could you explain a bit more - you probably do no mean this assignment, right?

templates/2k11/admin/entries.tpl
onli: Right. I do not see `$entry_data` anywhere, it was not assigned.

include/admin/images.inc.php
onli: I wasn't sure about this! It was my understanding that the `toggle_dir` set here is not evaluated at all, it does nothing. I might have missed where it's applied?

include/admin/images.inc.php
onli: That's a mixup on my side, this change is not supposed to be in this PR and it's' not made be me. I will revert it here. But iirc this change is a correction, this event is called two times with different parameters, this normalizes it.

templates/2k11/admin/plugin_config.tpl
onli: Right. This whole part caused a lot of PHP 8 errors, which made me notice the code was wrong in the first place. Separate commits in the future!

plugins/serendipity_event_spartacus/serendipity_event_spartacus.php
onli: It caused an error because $this->parser does not exist. And it can not exist, was my thought at time of removal.

include/functions_routing.inc.php
onli: I think this code never worked, $week would never be initialized. This must have been broken before.

include/functions_entries.inc.php
onli: Ah, the print. Right!

onli:
> Technically that should be possible, I'm not deep enough into Smarty these days to know, though.

Smarty has not released an release that claims PHP 8.0 support yet. Maybe this will be in it. But I'm not sure we should wait for it or patch Smarty more than absolutely needed?

include/plugin_api.inc.php
garvinhicking: Sigh, ok. I haven't tried that specifically yet, so even if we (before the language include) do some error_reporting() redirection we can't just load the undefined constants from a file? It would be a pain to change all language files to something like "if (!defined('LANG_STUFF')) define('LANG_STUFF', 'Content')"...

include/functions_config.inc.php
garvinhicking: I did not mean that specific line, but the one after. We have two $is_valid_user = false assignments now.

serendipity_config.inc.php
garvinhicking: This should then better be outsourced as a feature branch or something else - maybe at least something like a "TODO/FIXME: PHP8" line that can later be addressed?

include/functions_entries_admin.inc.php
garvinhicking: I meant the `url?serendipity['GET']['title']=<strong>evilHTML</strong>` possibility maybe, because the lines after this seem to pass through a GET variant (don't know if we really read from GET or if we just abused it as a superglobal to store a database title in)

templates/2k11/admin/entries.tpl
garvinhicking: Odd that this didn't lead to reports, but good that it can get fixed. But I'd isolate this as a specific bugfix commit (and document in NEWS)

include/admin/images.inc.php
garvinhicking: Hm ok, but I believe the proper code should contain the larger array, not just the one with "image_id"...?

plugins/serendipity_event_spartacus/serendipity_event_spartacus.php
garvinhicking: Oh true, probably it needs to use $p instead of $this->parser. Setting the proper target encoding doesn't seem unimportant to me... :D

include/functions_routing.inc.php
garvinhicking: I think in earlier codes (maybe before the change of proper routing was introcued) the $week parameter was possible to come from the URL directly `/archives/W43.html`...

include/admin/images.inc.php
onli: It really should not be in this PR. I will split this PR in a frontend/backend-PR, update smarty before that in master, also apply the UTF-8 fix there, and then re-apply my changes step by step. Then this change should vanish and the PRs be easier to review.

include/plugin_api.inc.php
onli: We in general don't need this. Given that everything renders it seems like we always redefine all the constants anyway, at least in the core and default plugins.

> so even if we (before the language include) do some error_reporting() redirection we can't just load the undefined constants from a file?

We might be able to hide all the Warnings for a moment with some voodoo, but that's magic to me. I couldn't build that hack, and think we shouldn't do it - would it work on all servers etc.?

include/functions_config.inc.php
onli: Hm. Wanted to explain that we need both, one for the first check, the second for the variable to be set back if it was true once. But no, if `$is_valid_user` is true once the loop will exit anyway (or rather, continue again and again, which should be a break probably). In this case we can make the continue a break and remove the second assignment.

templates/2k11/admin/entries.tpl
onli: Okay. But please not that it is also a PHP 8 compatibility fix, as the old code threw a warning there. This is one of the cases that is hard to separate.

plugins/serendipity_event_spartacus/serendipity_event_spartacus.php
onli: Worked before^^

include/functions_routing.inc.php
onli: I'd assume so as well, but I grepped for $week and nothing came up outside of .tpl files. Where was this assigned? Do you know enough about how this was supposed to work to restore it? I know nothing about the archive - but if we can we should of course fix in (in a separate PR).

onli:
Github is not recomputing the files changes overview. To toggle that for preparing two split PRs (one frontend, which I hope we can merge pretty immediately, and one backend that still needs more work) I will close here.

include/admin/images.inc.php
onli: Right, I missed it. Will be part of the new backend pr.

include/functions_config.inc.php
onli: Changed in the new PR

templates/2k11/admin/entries.tpl
onli: Separated into https://github.com/s9y/Serendipity/commit/e2682dbd79b6d37045226711a457ce544c27a395

templates/2k11/admin/plugin_config.tpl
onli: Separated into https://github.com/s9y/Serendipity/commit/ec70938faf6236dcc5f7b38c7c28373fd1200877

include/admin/entries.inc.php
onli: I think these variables are filtered above, at https://github.com/s9y/Serendipity/blob/20c0bed069af1a2a7cc815b9a433ad9d637b798d/include/admin/entries.inc.php#L228. And they also support just a limited set of inputs. But I might miss something. It would be great if someone could double-check this.

include/functions_entries_admin.inc.php
onli: Testing this, the title seems to be properly escaped when set into the form later on.

PR #758, opened 2021-04-25
Add config and dev dependency for starting to use phan

onli:
Phan enables a static code analysis and seems to be highly configurable. The goal is to have it enabled, to:

 1. At first identify additional PHP 8 compatibility issued
 2. Later run on PRs via Githubs CI infrastructure
 
This PR sets a minimal configuration that should identify some compatibility issues. Run it with:

```
bundled-libs/bin/phan --processes 3 --progress-bar -o analysis.txt
```

Please note: Installing phan included updating the PHP version in the composer.json and lead to lots of updates in bundled-libs. If we merge https://github.com/s9y/Serendipity/pull/757 first we can clean that up a bit. 

I'm still investigating alternative options myself.


onli:
@garvinhicking You preferred a different checker than phan, right? Should I go ahead and close here, or should I finish this up to merge it?

garvinhicking:
I think phan isn't the community's choice these days, phpstan is quite on top of things. However that one doesnt provide a phar, so we'd need a composer dependency for it

onli:
Let's do it a different way then, this PR would need a refresh anyway. I'll close here :)

Issue #128, opened 2014-04-24
jquery "Link" button has problematic "selection" functionality


garvinhicking:
In 1.7 using the plain editor (serendipity_editor.js) you can highlight a word, click on the link button and then enter the link that is wrapped around the highlighted word.

In 2.0, when inserting text without a selection it works. With a selection it seems to garble the selection and overwrite the link instead use the selected text.

Ref: http://board.s9y.org/viewtopic.php?f=3&t=19871

PR #765, opened 2021-06-12
Hide PHP warnings in production mode

onli:
Serendipity works already with PHP 8, probably even without the changes I made during the last weeks. They mainly fight against the many warnings PHP is throwing now. So far, when s9y throws a PHP warning one of two things happens:

 1. In development mode, detemined by the version number including alpha-beta or set manually, execution stops. A warning becomes a full blown crash.
 2. In production mode the code continues, but the warning is printed at the top of the page.

What this change does is to hide the warnings only in production mode.

There is no chance we will succeed with fixing every plugin and theme until the end of the year, so that with PHP 8 they do not throw warnings. Not without a lot more support for that project. This is a stopgap solution to users can upgrade anyway and most likely continue using their old plugins and themes.

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,3 +1,5 @@
+   * Hide more PHP warnings in production mode, to ease the migration
+     to PHP 8
    * Fix: Deleting a user was not possible
    * New images added via the ML will set loading="lazy", improving
      site performance for visitors (only if height and width known)
@@ -11,7 +13,7 @@
    * Change backend_image_add hook to always contain same structure
    * Split date and time input in editor into two input fields
    * Improve performance of the media library by caching the file list
-
+   
 Version 2.4-alpha2 ()
 ------------------------------------------------------------------------
    * Adds 'image_id' to event 'backend_image_add' in addData array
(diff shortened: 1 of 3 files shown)


onli:
Added multiple reviewers as this small change could be seen as a development decision, on how we go forward with PHP 8.

ddeimeke:
This works as well for me as solution for #128

ddeimeke:
This works as well for me as solution for https://github.com/s9y/additional_plugins/pull/128

onli:
The feedback by @ddeimeke contains some more information for us than obvious:

 1. I made him test the https://github.com/s9y/Serendipity/pull/765/commits/9828a56439eb096696ec221e96e3a40f662621c1 because in his blog, he ran in at least two plugins where the german language file was incomplete, staticpage and serendipity_event_typesetbuttons.
 2. While my dev environment is uberspace, he uses a different hoster with Ubuntu. That's an additional hint that the error_reporting workaround of the second commit is valid on sound Linux environments. And if that one works, the workaround in the first commit will work as well.

Still waiting for additional feedback here, but this is a good sign towards beta readiness.

Issue #766, opened 2021-07-01
Beta preparation for PHP 8 compatible version


onli:
In the last weeks, months now, commit after commit fixed code parts in the core that threw Warnings/Error with PHP 8. It is time to finish that work and release a beta version for those users that want to use PHP 8. 

But I think some things still need to be done:

- [x] The error handling needs to be remodeled: In [production mode](https://github.com/s9y/Serendipity/blob/8d8aa4ce6625a2e34172afe548d9a0ee8bce8b4d/serendipity_config.inc.php#L57-L60), show no warnings/errors at all if PHP can continue (Marked as done, but might need improvement)
- [x] Then, enable production mode for beta and RC versions. This will enable users to continue using old themes and plugins with PHP 8 now without having to wait for a stable release.
- [x] #768
- [x] All additional themes should run without errors in production mode, which means just fixing fatal errors.
- [ ] Ideally all plugins should work in production mode. The plugins that don't might have to go, if there is no maintainer. I already fixed a bunch of the more important ones, even in non-production mode.

If you can help, please install the current master with PHP 8, then install your favorite plugins or themes. If errors occur please fix them, or at least report them in this thread.

onli:
Production mode enabled for beta/rc versions in https://github.com/s9y/Serendipity/commit/fa69a788a820903ab519a6050cd0c71508ff6b28

onli:
I checked the recommended themes. There might still be warnings in subpages and previews, but at least the main pages work for me without warnings now.

surrim:
I'm still working on it. Till now I have
` 21 files changed, 115 insertions(+), 62 deletions(-)` - _Serendipity_
` 9 files changed, 24 insertions(+), 41 deletions(-)` - _additional_plugins_
... also affecting 2k11 theme and some core functions/plugins

onli:
I'm surprised that 2k11 and core plugins still have to be changed, but together with other plugins that's certainly possible. Don't hesitate to commit and push the changes already, in doubt into a branch that we can test and merge.

surrim:
Please take a look at https://github.com/s9y/Serendipity/pull/769 and https://github.com/s9y/additional_plugins/pull/131.
I tried to do my best, but some code is not really good to understand. Maybe a list of PHP8 ready plugins would be great. Some plugins are looking very old and not really maintained any more.
Maybe there is also a Smarty update for PHP8.

onli:
[...]

PR #769, opened 2021-07-08
Php8 fixes for #766

surrim:


surrim:
See https://github.com/s9y/Serendipity/issues/766

include/functions_entries.inc.php
```diff
@@ -1113,17 +1131,19 @@ function serendipity_printEntries($entries, $extended = 0, $preview = false, $sm
     if ($use_grouped_array === false) {
         // Use grouping by date (default)
         $dategroup = array();
-        for ($x = 0, $num_entries = count($entries); $x < $num_entries; $x++) {
-            if (!empty($entries[$x]['properties']['ep_is_sticky']) && serendipity_db_bool($entries[$x]['properties']['ep_is_sticky'])) {
-                $entries[$x]['is_sticky'] = true;
-                $key = 'sticky';
-            } else {
-                $key = date('Ymd', serendipity_serverOffsetHour($entries[$x]['timestamp']));
-            }
```
onli, line 1122: This is leading to an error in line 1336, where `$num_entries` is still referenced. The old code failed when an entry was marked as sticky?

onli, line 1122: This change leads to an error in line 1336, where `$num_entries` is still referenced. The old code produced a warning when an entry was stickied?

include/admin/importers/sunlog.inc.php
```diff
@@ -240,7 +240,7 @@ function import() {
         return true;
     }
 
-    function importCategories($parentid = 0, $new_parentid = 0, $sunlogdb) {
+    function importCategories($parentid, $new_parentid, $sunlogdb) { // default: $parentid = 0, $new_parentid = 0
```
onli, line 243: In the other situations where the parameter without a default trailed the others, I just set a new default. Just to consider as an alternative in the future - it will tend to break less.

include/admin/comments.inc.php
```diff
@@ -375,7 +375,7 @@
 $data['get']['filter'] = $serendipity['GET']['filter']; // don't trust {$smarty.get.vars} if not proofed, as we often change GET vars via serendipty['GET'] by runtime
 $data['commentReplied'] = false;
 
-if (!is_object($serendipity['smarty'])) {
+if (!is_object($serendipity['smarty'] ?? null)) {
```
onli, line 378: Did this happen to you, that smarty wasn't initialized yet at this point?

include/functions_routing.inc.php
```diff
@@ -37,33 +37,35 @@ function locateHiddenVariables($_args) {
...
-            $serendipity['head_subtitle'] .= SUMMARY . ' - ';
-            unset($_args[$k]);
-        } elseif ($v[0] == 'C') { /* category */
-            $cat = substr($v, 1);
-            if (is_numeric($cat)) {
-                $serendipity['GET']['category'] = $cat;
-                unset($_args[$k]);
-            }
-        }
+        if (strlen($v) > 0) {
```
onli, line 40: I'm surprised adding this was necessary - do you remember which situation this improves?

include/lang.inc.php
```diff
@@ -53,7 +53,7 @@
     // Needs to be included here because we need access to constant LANG_CHARSET definied in languages (not available for compat.inc.php)
 
     if (function_exists('mb_language')) {
-        @mb_language($serendipity['lang']);
+        @mb_language('uni');
```
onli, line 56: I had to look up why you changed that. Could you add a comment?

Also, did I get this right? The old code was wrong because only a limited set of languages is allowed here, so you set it to uni - for unicode- if possible?

onli (commented):
Looks very good! Just the $num_entries bug needs to be fixed (can you reproduce that error?), and the mb unicode change be explained. After that I'd like to merge this if you give the green light!

include/admin/comments.inc.php
surrim: This happens to me at many places. Finally I decided to search for the pattern instead of fixing only crashing code. When you have better code here, please paste me.

include/lang.inc.php
surrim: It was crashing here, because I'm (still) i Spain and `@mb_language('es');` crashed. I think it's not the best solution because there are also non-uniode installations. Maybe a switch would be great.

include/functions_entries.inc.php
surrim: Yes, because `$entries[$x]` only works when it's a normal array. Sometimes there are ´'id'´s in the array. Is `$num_entries` one less in that case?

include/functions_routing.inc.php
surrim: Yes, it happened when I visited `"https://bla.blah/archives/"`

include/functions_entries.inc.php
onli: `$num_entries` is simply not defined now. Or do you mean: Is it one less when there are ids in the array?

include/admin/comments.inc.php
onli: No, okay. As code it's fine, I was just surprised.

include/admin/importers/sunlog.inc.php
surrim: Do you want to set `null` here as default? There are also 4 other places where I fixed it like that.

include/functions_entries.inc.php
surrim: Yes, for example when you have
`array(0 => 42, 1 => 144000, 2 => -1/12, 'id' => 777)`
then
´$num_entries == 3´?

surrim:
> After that I'd like to merge this if you give the green light!

At the moment I'm updating 142 version numbers and ChangeLog's. Meanwhile you can test with all debugging output ;)

include/lang.inc.php
onli: Would be interesting to see what this fell back to before! Then we could switch it to that, or to the correct values on non-utf8 enabled blogs. But shouldn't block this PR I think.

include/functions_entries.inc.php
onli: I'll check that against the old code

include/admin/importers/sunlog.inc.php
onli: Yes, that might be a good idea (the importer need work anyway, but still)

include/functions_entries.inc.php
onli: Hm, I'm not able to reproduce that error. On my dev blog, `$entries` only has numeric keys. Do you remember in which situation exactly this failed?

surrim: No, I don't remember anymore. I fixed all errors that occurred.
Now I fixed this code with ´$num_entries = count($entries)`, subtracting one if ´id´ was set.

include/admin/importers/sunlog.inc.php
surrim: I changed the last parameter to ´null´.

PR #773, opened 2021-11-26
Fix missing mysqli_sql_exception in installer

### LINKED ISSUE #772: Uncaught mysqli_sql_exception in installer
Tried to set up a fresh S9y installation with PHP 8.1 and the MySQLi backend - which failed due with

`Fatal error: Uncaught mysqli_sql_exception: Table 'db.serendipity_authors' doesn't exist in /var/www/html/include/db/mysqli.inc.php on line 66`

An exception seems to be expected in the installer, but only a `PDOException`, not a `mysqli_sql_exception`:

https://github.com/s9y/Serendipity/blob/d79ea2742235f1410965b4bc45108c60de3f9ef7/include/admin/installer.inc.php#L327

We should probably add another catch block

```php
    } catch (mysqli_sql_exception $e) {
        $t = null;
```

mattsches:
Fixes #772

### DIFF
--- a/include/admin/installer.inc.php
+++ b/include/admin/installer.inc.php
@@ -326,6 +326,8 @@ if ( (int)($serendipity['GET']['step'] ?? null) == 0 ) {
         $data['authors_query'] = $t;
     } catch (PDOException $e) {
         $t = null;
+    } catch (mysqli_sql_exception $e) {
+        $t = null;
     }
 
     if ( is_array($t) ) {


onli:
@mattsches, will this work on a system that does not have mysql installed?

garvinhicking:
I'd also argue that this try/catch could better be performed directly within serendipity_db_query and return a straight false/null in case of an exception so the installer can continue?!

mattsches:
> @mattsches, will this work on a system that does not have mysql(i) installed?

Good question, I honestly don't know. I will try to figure this out, but it will take some time to set up a system without mysql.

Of course, Garvin has a point in that it's probably not ideal to check this in the installer, but I'm not sure I want to touch the `serendipity_db_query` functions because of this 😬

mattsches:
Short update: I managed to set up a container withoug `php-mysql` and tried an installation using postgres. The `catch` block was not an issue because no `mysqli_sql_exception` was thrown.

Interestingly, the `serendipity_db_query` in `postgres.inc.php` returns `false` for the statement and does not throw an exception, so that's a +1 for @garvinhicking 's suggestion, I guess.

I will take another look at it later on.

onli:
@mattsches Since this does work on a system without mysql I'd be fine with merging this as is for now, depending on whether you have the time to work on @garvinhicking's suggestion :)

onli:
@mattsches Should we go ahead?

onli:
Merged because this started to affect others, https://board.s9y.org/viewtopic.php?p=10454832.

PR #778, opened 2022-01-15
Improve permalink generation

### LINKED ISSUE #777: Duplicate permalink because of sanitisation
Similar to #738 I usually blog in English but also - rarely - in German. Instead of messing around with multilanguage, I've decided to just add a flag emoji (🇩🇪) to the title for those few German entries. E.g.

* Awesome Product
* Awesome Product 🇩🇪

Also, I don't want to bother users with the entry id, so I configured my permalinks to be `/article/YYYY-MM-DD-%lowertitle%.html`.

However, those emojis get stripped away when generating the permalink so both entries end up with the same permalink `/article/2020-02-02-awesome-product.html`. When opening the URL, the first result from the database wins - so depending on which entry I made first, visitors will ALWAYS see the English or the German version. There's no way for users to get the other.

I think there should be some check for collision and if so, the permalink should be made unique again - e.g. by adding `-1` to the end or similar. (Also, there maybe should be a `UNIQUE` constraint on the database table's field to make it impossible for such collisions to be added.)

Another way would be to make the sanitisation routine Unicode-aware. The flag emoji is composed of 2 [Regional Indicator Symbols](https://en.wikipedia.org/wiki/Regional_indicator_symbol), i.e. the special letters 🇩 and 🇪, and could be turned into the actual letters "DE".

mbirth:
This change brings 3 improvements to the permalink generation:

1. Wraps site-wide permalink changes into a transaction so it finishes much faster (esp. with SQLite database)
2. For languages where no replacements (`i18n_filename_*`) are defined, the default replacement happened AFTER all UTF-8 characters were stripped by the `utf8_decode()`. Which seemed pointless.
3. Introduces translation of flag emojis (e.g. 🇩🇪 - which actually consists of the 2 [Regional Indicator Symbols](https://en.wikipedia.org/wiki/Regional_indicator_symbol) 🇩 and 🇪) into latin letters (`DE` in this case) - Fixes #777 

(I had to convert the file to UTF-8 so I don't know what will happen if you have a site running on a non-utf-8 charset. It'll probably just not replace anything. But then, it's 2022 and UTF-8 should be default.)


onli:
This looks really good, thank you. Given that the master is currently in a beta phase, I'll make a note to test this in my dev blog and will try to come up with a test scenario, but from the looks of it this should just work.

onli:
Thank you [mbirth](https://github.com/mbirth)! Let's not forget to double check this feature before a stable release, but this was too clean to let it linger.

I will add a line to docs/NEWS, in the future it would be nice to have it as part of the PR :)

PR #785, opened 2022-08-09
Fix more cases of unset  (php8 compatibility)

hannob:
The logs show warnings of "Undefined variable $useAtom" for lines 215/218 in serendipity_plugin_syndication.php. This fixes these, it's basically the same as
https://github.com/s9y/Serendipity/commit/a3253a0d5512c484d72df973c32e7f46e810bb28
just for two more uses of this variable.

Alternatively one could rework the logic at the beginning to make sure $useAtom is set to False and never undefined.


onli:
Could you add an increment of the version number of the syndication plugin? It should be a 2.3.1 now.

onli:
Thanks!

PR #787, opened 2022-10-05
Encode square brackets in URL

hannob:
The W3C validator complains about the links containing [cview] (comment threading) in blog posts. This is due to the fact that [ and ] need to be url-encoded in href attributes. This PR fixes this.

See also this discussion on stackoverflow: https://stackoverflow.com/questions/40568/are-square-brackets-permitted-in-urls


onli (approved):
Tested and works

Issue #791, opened 2022-11-16
Uncaught ArgumentCountError: 4 arguments are required, 2 given in [path]/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php:1070


hannob:
I am seeing errors like this in my log file:
```
PHP Fatal error:  Uncaught ArgumentCountError: 4 arguments are required, 2 given in [path]/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php:1070
Stack trace:
#0 [path]/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php(1070): sprintf()
#1 [path]/include/plugin_api.inc.php(1188): serendipity_event_spamblock->event_hook()
#2 [path]/include/functions_comments.inc.php(1017): serendipity_plugin_api::hook_event()
#3 [path]/include/functions_trackbacks.inc.php(388): serendipity_saveComment()
#4 [path]/comment.php(108): add_trackback()
#5 {main}
  thrown in [path]/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php on line 1070
```

This is the code around the area:
```
                                if (!is_array($parts)) { // not a valid URL
                                    $this->log($logfile, $eventData['id'], $tipval_method, sprintf(PLUGIN_EVENT_SPAMBLOCK_REASON_IPVALIDATION, $addData['url'], '', ''), $addData);
                                    if ($trackback_ipvalidation_option == 'reject') {
                                        $eventData = array('allow_comments' => false);
                                        $serendipity['messagestack']['comments'][] = sprintf(PLUGIN_EVENT_SPAMBLOCK_REASON_IPVALIDATION, $addData['url']);
                                        return false;
                                    } else {
                                        $eventData['moderate_comments'] = true;
                                        $serendipity['csuccess']        = 'moderate';
                                        $serendipity['moderate_reason'] = sprintf(PLUGIN_EVENT_SPAMBLOCK_REASON_IPVALIDATION, $addData['url']);
                                    }
                                }
```

It looks to me the issue is with the sprintf calls. PLUGIN_EVENT_SPAMBLOCK_REASON_IPVALIDATION is a constant ('IP validation: %s [%s] != sender ip [%s]') with three %s string placeholders in it, so the sprintf call needs more parameters. As you can see in this code piece, one of the sprintf calls does it right by passing empty strings.

[...]

PR #792, opened 2022-11-16
Fixes call of sprintf with wrong number of arguments for error messag…

hannob:
…e. Also change to more meaningful error message.

Furthermore fix a logic bug. Before the code comparing the sender IP with the URL IP would be executed even if the URL parsing failed.

This fixes bug #791

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,6 +1,9 @@
 Version 2.4-beta2 ()
 ------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


onli:
The patch looks good and the old behaviour indeed like a logic bug. The plugin could use a cleanup.

Could you raise the plugin version to 1.89.4 number and add a comment to docs/NEWS?

hannob:
Ok like this?

PR #799, opened 2022-12-12
Host all of 2k11s webfonts locally

onli:
Limit potential for legal problems in the EU and specifically in Germany.

I used https://gwfh.mranftl.com/fonts to download all the webfonts of 2k11s config (Droid Serif is now Noto Serif) and changed the **config.inc.php** to emit the needed CSS.

It would have been nicer to have that CSS code in regular CSS files, but the path substitution was a problem for that option.

Looking for feedback here, but in doubt I'd merge this soonish to make it part of 2.4.1.


onli:
@fe-hicking :)

Issue #784, opened 2022-08-07
function strftime is deprecated (php 8.1)


hannob:
With PHP 8.1 I'm getting these warnings in the log:
```
PHP Deprecated:  Function strftime() is deprecated in [path]/include/functions.inc.php on line 143
```

The use of the strftime function is discouraged and will likely be removed in future PHP versions:
https://www.php.net/strftime

garvinhicking:
Sadly don't have much time these days, but I'd suggest to use a polyfill to make it work again. Adjusting to the immutalbe DateTime Interface for legacy application is a major pain.

Example for a polyfill: https://bestofphp.com/repo/alphp-strftime

That could be copy+pasted to serendipity_compat.inc.php or something like that, and be a drop-in replacement...? @onli

onli:
Using the polyfill sounds like a good idea to me!

surrim:
> Sadly don't have much time these days, but I'd suggest to use a polyfill to make it work again. Adjusting to the immutalbe DateTime Interface for legacy application is a major pain.

It seems it's quite a lot of work. There is another format for the [date()](https://www.php.net/manual/en/function.date.php) function, no way to reuse the format strings in all translations without a polyfill.

surrim:
> Example for a polyfill: https://bestofphp.com/repo/alphp-strftime
> 
> That could be copy+pasted to serendipity_compat.inc.php or something like that, and be a drop-in replacement...? @onli

Works fine! 💃‍
I implemented that [here](https://github.com/s9y/Serendipity/commit/a404235e06df0ad4495c917b01cfda1aeaa88395#diff-4bc07f5b555f1123057ec3e3e67faa582997401895813e1c1468e8886e9411f6R143), but it's not ready for merging because of other bugs.

onli:
I added a check to the installer. Given that 2.5 was already released without further documentation about this change I think it's okay to leave it like this.

PR #802, opened 2023-02-10
Php82 fixes

surrim:
I fixed the most critical things in order to get S9Y working on PHP8.2.

I had to update Smarty to [v4.3.0](https://github.com/smarty-php/smarty/releases/tag/v4.3.0) in order to have PHP8.2 support. Without that the screen was full of warnings and critical errors. I included the package with composer, so I also had to rename some paths in the code.
I also included a strftime polyfill which seems to work great, see https://github.com/s9y/Serendipity/issues/784

Can we merge the improvements already made in already?
At least it is a good starting point to fix warnings because it is no longer completely broken.


bundled-libs/psr/simple-cache/composer.json
```diff
@@ -6,11 +6,11 @@
     "authors": [
         {
             "name": "PHP-FIG",
-            "homepage": "http://www.php-fig.org/"
+            "homepage": "https://www.php-fig.org/"
         }
     ],
     "require": {
-        "php": ">=5.3.0"
```
onli, line 13: This change is strange. I see that it was not intended, but how can we solve it? voku/simple-cache has psr/simple-cache as dependency, but sets it to 1.0 || 2.0, with only 2.0 requiring PHP 8.0. 

We should either bundle 1.0 (if compatible with PHP 8), somehow bundle both and use the one that fits, or swap out voku/simple-cache?

stevleibelt:
One general question, why merge to master and not to a feature branch like "update-to-php-8-2"?

To much complexity overhead?

Cheers and thanks for your work.

surrim:
Because I've changed many files and don't want to have big merge conflicts. Nothing is rewritten, so now it's working. If somebody else is making changes, maybe the code is failing later.

bundled-libs/psr/simple-cache/composer.json
surrim: I have no idea, I don't know any of them. I have only updated Smarty and included the strftime module - this was the result.
Fell free to modify the branch ;-)

onli:
@surrim, should we merge this and work on the remaining tasks - like the cache library - in master?

PR #807, opened 2023-06-20
Style the RSS feed with an XSLT

onli:
This adds the pretty-feed-v3.xsl from https://github.com/genmon/aboutfeeds to the RSS page. It does so via the rss.php, as something in our code breaks the xslt line when it is added to the `.tpl` directly, the `?>` gets removed. 

As a result, instead of raw xml code the user sees this when clicking on the rss link:

Screenshot 2023-06-20 at 17-45-13 onli blogging (Login onli) Web Feed

Do not merge this before https://github.com/genmon/aboutfeeds/issues/34 is solved, the license is incompatible. I assume this is an accident, if not we would have to write  a new .xslt.


onli:
I assume the base project is not interested in addressing the license issue. I'll give this up then.

PR #810, opened 2023-08-18
Fixed Russian translation

varakh:
I fixed Russian translation of existing files and added translation where it was missing.


onli:
Wow @varakh, that's great. Thank you.

The changes under htmlarea/ are a bit problematic. That's basically an external dependency we have inserted, when it is updated the changes would be lost. The right place for that commits would be https://ckeditor.com/. But that does not need to block this PR.

All the other changes will persist, the on to the core lang file and the plugin lang files. So I'll go ahead and merge this.

varakh:
@onli, thank you  for the information about ckeditor.

Issue #821, opened 2024-02-14
Serendipity Composer Update | Uniify plugins in core and additional_plugins


surrim:
There are four plugins which exists in Serendipity and additional_plugins:

| Plugin name                    | S9Y    | AP     |
| ------------------------------ | ------ | ------ |
| serendipity_event_gravatar     | 1.63.1 | 1.62.3 |
| serendipity_event_multilingual | 2.36.2 | 2.35.1 |
| serendipity_plugin_authors     | 2.2    | 2.2    |
| serendipity_plugin_quicksearch | 1.2    | 1.2    |

There should be only one source. Everything which is optional should be in additional_plugins.

garvinhicking:
I believe the reasoning at some point was to "bundle" important plugins with our core release. The better way would've been to do this via on-release bundling - which we weren't able to do "back in the days". Also I think we did some things with the plugins that affected their core compatibility, but I'm not sure of this.

A problem is that when we now move plugins outside of the core, they'll not get deleted when people update.

So maybe the best course of action is to "just" synchronize the affected plugins?

The ideal way of course would be to make plugins be composer packages, migrate the core to use composer for plugins too, and then create a proper release architecture.... I did some initial work for a composer move (see the git branch), but I can't do it alone and need some motivation (&time). 🥹

stevleibelt:
@garvinhicking Is there a list or "howto"-Section you can point volunteers to support you on the composer task?

garvinhicking:
@stevleibelt That would be terriffic, and might get me involved somehow. It's just so much that I'm seeing a steep hill I'd have to climb alone. Serendipity has a very small userbase and I'm just not sure there is "public demand" for that. But the task at hand would be a cool one, and we'd all learn a lot from it. So... if there are volunteers (LIKE YOU? :-)) then please do get in touch.

I'm still on the s9y.slack.com, but the more transparently public we can communicate this, the better.

I've just updated https://github.com/s9y/Serendipity/blob/feature/composer/README.markdown#concept again to add some thoughts...

( @onli )

onli:
@garvinhicking Just to make sure you have seen it, @mmitch also volunteered to help at https://www.onli-blogging.de/2323/comments.html#c10436

onli:
[...]

PR #823, opened 2024-02-18
add simple CI pipeline with basic PHP test runner

mmitch:
(ping @garvinhicking ref. https://www.onli-blogging.de/2323/15-Jahre-Serendipity-als-Entwickler-ein-Rueckblick-und-ein-Ausblick.html#c10436)

I have prepared a simple GitHub Action CI pipeline that runs the basic PHP tests on every push and will annotate commits and Pull Requests with the build status.

The tests run for all current PHP versions (that means 8.1, 8.2 and 8.3). The matrix can be expanded to other versions as well, but I don't know if we need to test against deprecated PHP versions without security support. On the other hand, nightly builds of PHP 8.4 could also be used as a test target.

To make the tests pass I had to update them to make them work with a current version of PHPUnit.

Things currently not included:
* There is no build status badge for `README.markdown` yet. If you like to have one, I'll add it :-)
* Code coverage is not computed yet. With the low count of test cases this does not seem too useful at the moment.
* `FrontendTest.php` is not included yet. The test framework _Selenium RC_ is triple-deprecated (Selenium RC is Selenium 1; Selenium 1, 2 and 3 are already deprecated; Selenium 4 is current) and I can't even find downloads of the mentioned `selenium-server.jar` (Selenium homepage and GitHub project offer Selenium 2.39.0 as oldest download). I'll see if I can update the tests to work with a current version somehow or if it is feasible to switch to another test framework (any wishes?) - but not as part of this PR ;-)

### DIFF
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -33,8 +33,7 @@ jobs:
         coverage: none
       env:
         fail-fast: true
-    - name: Check that PHP works
-      shell: php {0}
+    - name: Run Tests
       run: |
-        <?php
-        echo "Hello world from PHP " . PHP_VERSION;    
+        cd tests
+        phpunit --configuration config.xml.dist --coverage-html report include/*
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -1,0 +1,40 @@
+# This workflow will run the tests on various versions of PHP.
+
+name: PHP Tests
+
+on:
+  push:
+  pull_request:
+    types: [opened]
+# TODO: enable to run once a week regardless of commits, but this will expire if repo is inactive for a longer time
+#  schedule:
+#    - cron: '12 3 4 * *'
+
+jobs:
+  test-php:
+    name: Run PHP Tests
+
+    runs-on: ubuntu-22.04
+
+    # for available versions see https://github.com/marketplace/actions/setup-php-action#tada-php-support
+    strategy:
+      matrix:
+        phpversion: ['8.1', '8.2', '8.3']
+
+    steps:
+    - uses: actions/checkout@v4
+    - name: Setup PHP and dependencies
+      uses: shivammathur/setup-php@v2
+      with:
+        php-version: ${{ matrix.phpversion }}
+        extensions: pdo-sqlite
+        tools: phpunit
+        # TODO: enable coverage check
+        coverage: none
+      env:
+        fail-fast: true
+    - name: Check that PHP works
+      shell: php {0}
+      run: |
+        <?php
+        echo "Hello world from PHP " . PHP_VERSION;    
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -19,7 +19,7 @@ jobs:
     # for available versions see https://github.com/marketplace/actions/setup-php-action#tada-php-support
     strategy:
       matrix:
-        phpversion: ['8.1', '8.2', '8.3']
+        phpversion: ['7.4', '8.1', '8.2', '8.3']
 
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/php_tests.yml
+++ b/.github/workflows/php_tests.yml
@@ -36,4 +36,4 @@ jobs:
     - name: Run Tests
       run: |
         cd tests
-        phpunit --configuration config.xml.dist --coverage-html report include/*
+        phpunit --configuration config.xml.dist include/*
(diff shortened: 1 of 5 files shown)


garvinhicking:
Wow, that's pretty cool. I wasn't aware the tests we have are really anything at this point, but happy to see the can still act as a foundation.

Have you by chance seen the "feature/composer" branch and https://github.com/s9y/Serendipity/issues/821#issuecomment-1948542033?

Would you be interested in aligning this PR with that branch and maybe work on it together?

(Otherwise I'd happily merge this PR into main, it's an improvement either way. If we could add PHP 7.4 into the matrix mix, I'm sure @onli would be happy, because that's still used in some Linux distro LTS's)

Many, many, many thanks for getting involved, this puts a smile to my face.

garvinhicking (approved):

onli:
> (Otherwise I'd happily merge this PR into main, it's an improvement either way. If we could add PHP 7.4 into the matrix mix, I'm sure @onli would be happy, because that's still used in some Linux distro LTS's)

Exactly right :)

mmitch:
The tests run under PHP 7.4 , too, I've just added it to the build matrix.

I think I'll first try to tackle the frontend testsuite before looking at #821 in depth – I've never worked with composer (and I don't want to use Docker :wink:), so I'll have to dive a bit deeper before doing anything.
With regard to my first glance at https://github.com/s9y/Serendipity/blob/feature/composer/README.markdown#concept
* Linting and static code analysis should be possible if there are any PHP Actions available (GitHub CodeQL does _not_ seem to support PHP, or I would already have added it here)
* Dependency checks should also be possible if there are any Actions for that. Perhaps it is already covered by GitHub itself (there is a dependency management, but the languages that I typically use aren't supported).
* Release and asset building might be hardest because I've never really done that and I don't know anything about PHP release artefacts and packaging.

Feel free to merge this in the meantime, I can use it as a foundation if I take a look at the composer-feature.

garvinhicking:
Hi!

Thanks a lot, I just merged your PR into both branches (main and feature/composer).

> I think I'll first try to tackle the frontend testsuite before looking at #821 in depth – I've never worked with composer (and I don't want to use Docker 😉), so I'll have to dive a bit deeper before doing anything. With regard to my first glance at https://github.com/s9y/Serendipity/blob/feature/composer/README.markdown#concept

You should be able to help with the "composer move" effort even without too much composer knowledge I hope.

With Docker and ddev you would be setup in a minute, just git clone the `feature/composer` directory, then `ddev start`, `ddev composer install` and then `ddev launch` and you would be good to go with a completely distinct webserver environment with PHP+MySQL.

Without that, it's a lot more configuration of course:

* Create a directory like /var/www/html/serendipity (where you store your webserver files)
* GIT clone the `feature/composer` branch into that directory
* Setup your webserver for a distinct virtualhost, map the DocumentRoot to /var/www/html/serendipity/public
* Inside the /var/www/html/serendipity directory, do a `composer install`
* Then you should be able to point your browser to your virtualhost and see the output

Of course currently the branch is non-functional, it throws errors because the PHP include directories and paths are not referenced properly, because I moved around directories but did not change the code for it. That's the "busywork" we need to work on :-D 

> * Linting and static code analysis should be possible if there are any PHP Actions available (GitHub CodeQL does _not_ seem to support PHP, or I would already have added it here)

I'd do that with phpstan and php-cs-fixer, there are .github workflows out there I utilize, e.g. https://github.com/TYPO3-Documentation/render-guides/blob/main/.github/workflows/main.yaml - it should be straight-forward to port that to serendipity, once we use composer.

> * Dependency checks should also be possible if there are any Actions for that. Perhaps it is already covered by GitHub itself (there is a dependency management, but the languages that I typically use aren't supported).

There's some GitHub action workflows available that use dependabot, I do have some bootstrap code for that (https://github.com/TYPO3-Documentation/render-guides/blob/main/.github/dependabot.yml) which should also be easy to port to Serendipity, again - once composer is properly utilized :-) 

> * Release and asset building might be hardest because I've never really done that and I don't know anything about PHP release artefacts and packaging.

I've got some experience with that too, steps would be similar to something like https://github.com/sergeyklay/gh-actions-php-phar-create-release-example/blob/master/.github/workflows/release.yml

Also the creation of a docker container shouldn't be hard, with something like https://github.com/garvinhicking/typo3-documentation-browsersync/blob/main/.github/workflows/docker.yaml

Crossing my fingers I can get you hooked 👯‍♂️

onli:
> I'd do that with phpstan and php-cs-fixer

Just to note, IIRC I did have no luck with phpstan, it was not compatible with the oldschool serendipity PHP code. That is why I set on phan instead, at https://github.com/s9y/Serendipity/pull/758 is still the config.

So if you get phpstan to work we can close #758, otherwise maybe the config there can be re-used :)

PR #826, opened 2024-03-07
bugfix: Made comment token clean-up work in PostgreSQL also

HQJaTu:
SQL-table _options_ has column _name_ defined as varchar.
https://github.com/s9y/Serendipity/blob/master/sql/db.sql#L182

In code, the cleanup calculates an unix timestamp pointing to history one week ago. This interger is then compared with the mentioned varchar in SQL. This comparison might fly on some RDMSes, on PostgreSQL typing is strict and an int cannot be compared with a varchar.

This commit fixes the problem making comment handling working again.

### DIFF
--- a/include/db/db.inc.php
+++ b/include/db/db.inc.php
@@ -183,4 +183,21 @@ function serendipity_db_implode($string, &$array, $type = 'int') {
     return $string;
 }
 
+/**
+ * @access public
+ * @param   string Database table column name
+ * @param   string Database column type
+ * @return  string Column CAST() to chosen database
+ */
+function serendipity_db_cast($columnName, $type) {
+    global $serendipity;
+
+    if (stristr($serendipity['dbType'], 'sqlite')) {
+        return $columnName;
+    }
+
+    // Adds explicits casting for ANSI SQL -compliant DBs, like mysql and postgresql.
+    return "cast($columnName as $type)";
+}
+
 /* vim: set sts=4 ts=4 expandtab : */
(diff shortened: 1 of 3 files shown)


garvinhicking:
Nice catch, thank your for reporting! Let me check the codebase if we might have a similar problem somewhere else, I'm thinking about other timestamp-y fields used for login validation. Will get back to you on that tomorrow!

garvinhicking:
@HQJaTu I've found this in `include/functions_comments.inc.php` (serendipity_commentSubscriptionConfirm):


```
    if (stristr($serendipity['dbType'], 'sqlite')) {
        $cast = "name";
    } else {
        // Adds explicits casting for mysql, postgresql and others.
        $cast = "cast(name as integer)";
    }

    serendipity_db_query("DELETE FROM {$serendipity['dbPrefix']}options
                                WHERE okey LIKE 'commentsub_%' AND $cast < (" . (time() - 1814400) . ")");

```

This is also used similiary in `include/functions_config.inc.php` for `serendipity_issueAutologin()` and `serendipity_checkAutologin()`.

I wonder if, for consistency, you could alter your PR/patch so that we also use the explicit cast here, too? What do you think?

Best regards,
Garvin

HQJaTu:
Sure thing. I went that way.

As D.R.Y. -principle applies, I don''t repeat myself. I added a function to assist with casting, so it won''t be done in three separate places.

garvinhicking:
Nice!! I was secretly hoping you would see the uglyness in this 🤓

Thanks, I'll review properly next week and merge then. Looking good!

garvinhicking:
Would you maybe like to also use your CAST unification in the place `include/functions_config.inc.php` for serendipity_issueAutologin() and serendipity_checkAutologin() that I mentioned? :-)

garvinhicking:
Perfect! Thanks!

PR #827, opened 2024-03-13
bugfix: Tooltips working again

HQJaTu:
Admin-page tooltips were not doing anything when clicked.

Fix was to add a missing space between two accessibility HTML-attributes. Having HTML in working order enabled the functionality again.


garvinhicking:
Oooh. That is embarassing. 😅

Would you like to ammend your commit with an entry to docs/NEWS? That's our "living" changelog, you could attribute your fix there and make my merge process a bit easier.

Thanks!!

HQJaTu:
Again, hopefully my update is in the correct spot. Generally speaking, I have no idea on how release management works in this project.

garvinhicking:
Thanks!

Release management basically is:

* Whenever we make a commit, we also edit `docs/NEWS` and add a changelog entry there (unless it's a very basic change that doesn't need an entry).
* So for Pull requests, ideally if people want to write a changelog entry for that too, they could include it in their commit and we merge it directly. Then we maintainers don't need a distinc extra commit for it

Once we feel that there are important issues fixed or new features added, we create a new release and increase version numbers. Some info on that is here: https://github.com/s9y/Serendipity/wiki/Creating-a-release

Issue #828, opened 2024-03-16
serendipity_strftime() does not convert date to local charset


F3lisCatus:
My blog is set to using ISO 8859-1 instead of UTF-8 for historical reasons. After updating to 2.5.0 and PHP 8.2 I noticed the month on top of my entries was being displayed as a UTF-8 string ("MÃ¤rz" instead of "März"). After some fiddling around with the source code and a short discussion on the forums (https://board.s9y.org/viewtopic.php?t=26376) I ended up adding the following line to /include/functions.inc.php:

```
# diff functions.inc.php functions.inc.php.orig 
163d162
<     if (LANG_CHARSET != "UTF-8") $out = mb_convert_encoding($out, LANG_CHARSET, 'UTF-8');
```

..which seems to work fine for me.

onli:
@F3lisCatus With the PR stalled and suggesting that the conversion was not necessarily the right solution, I assume this can't be reproduced anymore or is okay for now. I will close here, please re-open this issue if the problem reappears.

PR #829, opened 2024-03-16
Update functions.inc.php

F3lisCatus:
If charset in use is not UTF-8, convert function $out to LANG_CHARSET first

### DIFF
--- a/include/functions.inc.php
+++ b/include/functions.inc.php
@@ -160,6 +160,7 @@ function serendipity_strftime($format, $timestamp = null, $useOffset = true, $us
         $out = mb_convert_encoding($out, 'UTF-8', 'ISO-8859-1');
     }
 
+    if (LANG_CHARSET != "UTF-8") $out = mb_convert_encoding($out, LANG_CHARSET, 'UTF-8');
     return $out;
 }
 


F3lisCatus:
See also issue 828: https://github.com/s9y/Serendipity/issues/828

garvinhicking:
I believe your problem is that you dont have a iso88591-locale installed on your server, so a UTF-8 locale is used instead...?

garvinhicking:
Here in this file we declare the used locales:

https://github.com/s9y/Serendipity/blob/1e7edd45353d6dba0a895f1aa68bc41b5c162a7c/lang/serendipity_lang_de.inc.php#L9

Your locale does not seem listed here; you could try to add it there and if it works we can make a PR for that change instead?

Thanks for helping to improve Serendipity :)

F3lisCatus:
The complete list of supported locales is quite extensive, here are the ones for German:

`de_DE
de_DE.iso88591
de_DE.iso885915@euro
de_DE.utf8
de_DE@euro`

Seems like my provider omitted the dash between 8859 and 1, could that be the problem?

F3lisCatus:
I edited /lang/serendipity_lang_de.inc.php line 9 to read `@define('DATE_LOCALES', 'de_DE.iso88591 de_DE.iso885915@euro, de_DE.ISO-8859-1, de_DE.ISO8859-1, german, de_DE, de_DE@euro, de');` and removed my line 162 from functions.inc.php, but does not solve my problem. "MÃ¤rz" is back! ;-) I tried some other variations but no luck.

I think I will just keep my patch in the functions.inc.php for now. I guess there is something wrong with my configuration. I always seem to run into localization problems anyways, it's kind of my thing. ;-)

garvinhicking:
If you want, let us debug this. I always want to know the root of this 😅
I'll get back to you next week, with a dummy script we should be able to debug this, if you want?

F3lisCatus:
Ok, but on weekdays I don't have that much time so beware of my response times. ;-)

onli:
@F3lisCatus That was a long response time ;) Did you have any luck with this, will you still pick this up?

onli:
I will close here, seems like progress staled. Please re-open or open a new PR if there is still a fix to be added to Serendipity :)

PR #835, opened 2024-04-03
Improve thumbnail quality and size

onli:
This tackles a couple of issues with the thumbnail creation:

 1. Gives jpeg thumbnails created without imagemagick a reasonable quality setting, so they are no longer as huge as they were before (often bigger than the full sized original).
 1. Tunes imagemagicks parameters suggested in **serendipity_config** and actually enables those settings, so no longer just the defaults are used.
 2. Finally, adds https://github.com/spatie/image-optimizer to apply image optimizers to the thumbnails, if they are installed.

That last step adds a couple of dependencies, plus the project has a security issue open at https://github.com/spatie/image-optimizer/issues/210. We should wait with this PR until after that one is patched, and I'm also okay if we just revert the last commit and forego that optimizer step. It might better fit as a plugin anyway. But I wanted to present the option. What do you think?


garvinhicking:
Interesting... the imageoptimizer looks like a huge dependency set.

Where exactly is it utilized? I could only see the "use" statement in functions_images, but how would that affect things?

onli:
> Ok saw it now, github didnt load completely due to diff size. Hm, I'm torn. 

Me as well.

> Do you have an example image where the optimizing effects show a real impact?

Yes, https://onli2.uber.space/s9y_dev/index.php?/archives/16-png-without-image-optimizer.html vs https://onli2.uber.space/s9y_dev/index.php?/archives/15-imageoptimizer-chain-png.html (make it a small browser window so it loads the thumbnail).  1.3MB vs 900KB. But it's a pathological case, one normally would not use .png for such a motive. And for jpg the difference was miniscule.

Probably best to revert the last commit, I think more and more this should be a plugin.

garvinhicking:
Agreed. Let's take out the dependency and better factor it into a plugin.

Thanks!

PR #842, opened 2024-06-06
Block .inc files in .htaccess template

onli:
Blocking .inc files might be seen as a security feature, for webserver configurations that execute files with that ending.


garvinhicking (approved):
👍

Issue #844, opened 2024-07-10
Unhandled mysqli exception in search with invalid search strings


hannob:
I'm monitoring my PHP error logs, which often helps me identify bugs in PHP applications.

Since a while, I'm regularly seeing unhandled mysqli exceptions by serendipity, like this:
```
PHP Fatal error:  Uncaught mysqli_sql_exception: syntax error, unexpected '-' in [path]/include/db/mysqli.inc.php:68
Stack trace:
#0 [path]/include/db/mysqli.inc.php(68): mysqli_query()
#1 [path]/include/functions_entries.inc.php(953): serendipity_db_query()
#2 [path]/include/genpage.inc.php(61): serendipity_searchEntries()
#3 [path]/include/functions_routing.inc.php(199): include('...')
#4 [path]/index.php(93): serveSearch()
#5 {main}
  thrown in [path]/include/db/mysqli.inc.php on line 68
```

What happens is that the search function of s9y can pass invalid search strings to mysql, and that causes this exception.
You can also test this on the main s9y blog, by passing a search string like "test -" into the search form:
https://blog.s9y.org/index.php?serendipity%5Baction%5D=search&serendipity%5BsearchTerm%5D=test+-&serendipity%5BsearchButton%5D=Los%21

This causes an error 500 for the user.

I think this should be caught somewhere and the user should get some form of error. I've looked a bit where to do that, but haven't yet written a patch, as it's not entirely obvious where to best handle this. The code in serendipity_searchEntries() is generic for various DB backends, but serendipity_db_query() doesn't really know that the query is a search, which could cause such errors.

Reporting it here, so maybe others can have a look.

onli:
Probably something to catch in https://github.com/s9y/Serendipity/blob/71f2c2dfaed766ed4a33096aed3524794859c0d3/include/genpage.inc.php#L61

onli:
This should be fixed now in current master, thanks to https://github.com/s9y/Serendipity/pull/846 by @GuillaumeValadas 

Thanks for the report @hannob ! I'll close here already, we can re-open if the issue remains or returns.

hannob:
Interestingly, I am still seeing such an error (longer stack trace this time) when applying this patch, but only on one of my s9y installations...
I'm not sure what can cause the different behavior... Any pointers appreciated.
The installation were I still see it is older (i.e. upgraded from older s9y versions, possibly something different in the mysql db?), and has a custom theme.

Error new:
```
[...]

PR #846, opened 2024-07-16
Fix search issue with special characters, and escape them in SQL. #844

GuillaumeValadas:
I write a fix to escape specials char and escape them in SQL to allow research without 500 server error.

Its related to the issue #844 

Hope it will help.

### DIFF
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -884,7 +884,7 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $relevance_enabled = true;
         if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
             $term = serendipity_db_escape_string($term);
-	    $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
+            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
         } else {
             $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term')";
         }
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -882,8 +882,11 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $cond['distinct'] = '';
         $term             = str_replace('&quot;', '"', $term);
         $relevance_enabled = true;
-        $term = serendipity_db_escape_string($term);
-        $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
+        if (preg_match('@([\+\-\*~<>\(\)"])\s*(\b\w+\b)@', $term)) {
+            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term' IN BOOLEAN MODE)";
+        } else {
+            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term')";
+        }
     }
 
     switch($serendipity['searchsort']) {
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -883,7 +883,8 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $term             = str_replace('&quot;', '"', $term);
         $relevance_enabled = true;
         if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
-            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term' IN BOOLEAN MODE)";
+            $term = mysqli_real_escape_string($serendipity['dbConn'], $term);
+	    $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
         } else {
             $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term')";
         }
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -882,12 +882,8 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $cond['distinct'] = '';
         $term             = str_replace('&quot;', '"', $term);
         $relevance_enabled = true;
-        if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
-            $term = serendipity_db_escape_string($term);
-            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
-        } else {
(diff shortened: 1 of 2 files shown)


include/functions_entries.inc.php
```diff
@@ -883,7 +883,8 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $term             = str_replace('&quot;', '"', $term);
         $relevance_enabled = true;
         if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
-            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term' IN BOOLEAN MODE)";
+            $term = mysqli_real_escape_string($serendipity['dbConn'], $term);
```
fe-hicking, line 886:
```suggestion
            $term = serendipity_db_escape_string($term);
```

include/functions_entries.inc.php
```diff
@@ -883,7 +883,8 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
...
+	    $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
```
fe-hicking, line 887:
```suggestion
            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
```

fe-hicking (commented):
I think this looks promising! @hannob @onli -- thoughts?

Thanks so much for providing a patch! 👏

include/functions_entries.inc.php
```diff
@@ -883,7 +883,8 @@ function &serendipity_searchEntries($term, $limit = '', $searchresults = '') {
         $term             = str_replace('&quot;', '"', $term);
         $relevance_enabled = true;
         if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
-            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term' IN BOOLEAN MODE)";
+            $term = serendipity_db_escape_string($term);
+            $cond['find_part'] = "MATCH(title,body,extended) AGAINST('\"$term\"' IN BOOLEAN MODE)";
         } else {
             $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term')";
```
garvinhicking, line 889: I wonder, we should use db_escape_string here too?

GuillaumeValadas: Maybe that mean you can remove if cond if you apply it anyway.

garvinhicking:
Thanks looks good to me!waiting for @hannob and @onli a bit to see if they approve too :)

onli:
I'm not sure the approach will work, but I might have misunderstood the issue. The example string failing was with a `-` at the end, and I thought this fails because the boolean mode expects a negated search term after that sign. I don't think that `mysqli_escape_string` (what `serendipity_db_escape_string` would call here) changes the -, so how does escaping the search term fixes the problem?

Also, the old code had a switch there between boolean and regular mode, won't that still be needed?

I absolutely might miss something here :) But I'd suggest to check whether the search still fully works with this patch applied.

garvinhicking:
> I'm not sure the approach will work, but I might have misunderstood the issue. The example string failing was with a `-` at the end, and I thought this fails because the boolean mode expects a negated search term after that sign. I don't think that `mysqli_escape_string` (what `serendipity_db_escape_string` would call here) changes the -, so how does escaping the search term fixes the problem?
> 
> Also, the old code had a switch there between boolean and regular mode, won't that still be needed? Wasn't the regular `against` a natural language search with theoretically better results?
> 
> I absolutely might miss something here :) But I'd suggest to check whether the search still fully works with this patch applied.

Oohh. Right. I see it now - the patch uses "" quotes around the term. So it actually removes the "BOOLEAN MODE" functionality for exact matching.

So we need something like:

Search input A: This is a long search query
Search input B: Some +keyword -badkeyword search
Search input C: Some malformed query +++++-

Input A should match "This is a long search query" without boolean mode.
Input B should match the term in boolean mode: "Some +keyword -badkeyword search" (so not for '"Some +keyword -badkeyword search"')
Input C should be caught. Probably we would need some kind of try/catch surrounding the query to catch a malformed query, and then probably remove all special characters from the term (the ones matched in `preg_match()`) and then re-try the query...

(Disclaimer: I haven't checked the initial bugreport thoroughly)

garvinhicking:
Thanks for the further refinement - that looks promising, I couldn't test various scenarious myself though. Can you tell which examples you used where it works and where it broke before?

I still feel a bit like this can still lead to cases where "BOOLEAN MODE" would be invalid, but I doubt we can hack together a more solid parser on our own...

GuillaumeValadas:
So now you enter boolean mode only with {operator}whatever.

no more "test -" goes to boolean mode anymore it goes on "normal" mode that cause no problems.

Feel free to test or comment if its not enough.

onli:
Oh, that's a good idea!

include/functions_entries.inc.php
garvinhicking: I think `@@[\+\-\*~<>\(\)"].[\S]*@` is currently wrong there, maybe a copy+paste mistake?

Anyhow... I'm afraid we regular-expression based evaluation of this might not bring us further. Because the regexp would match true on a string like: "+valid +valid +".

So I think we have two options here:

A: Introduce a tokenizer that analyzes the whole string and ALL parts in it, and performs checks on it
B: Execute the query and check if it works or not

Approach A will be quite complex and way too much effort to do this reliably. Not sure there is a parser/tokenizer for this out there, maybe part of Doctrine DBAL.

Approach B is a bit nasty but might be little effort. We could execute the query, use the "expect error" flag, and check if there is a result, something like:

```
$searchMode = 'non-boolean';
if (preg_match('@["\+\-\*~<>\(\)]+@', $term)) {
$termRows = serendipity_db_query("SELECT * FROM serendipity_entries WHERE $term LIMIT 1", true, 'both', false, false, false, true); // Last "true" is important here
  if ($termRows !== false) {
    $searchMode = 'boolean'; // Terms works as boolean expression
  }  
}
if ($searchMode === 'bolean') {
    $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term' IN BOOLEAN MODE)";
} else {
    $cond['find_part'] = "MATCH(title,body,extended) AGAINST('$term')";
}
```

(untested, just a thought)

GuillaumeValadas:
image

I add another regular expression check to not enter the boolean mode if any operator live alone at the end of the string search.

Any blue match in the image **don't go** in boolean mode.

I think now all cases are cover.
Maybe you can test this on your side.

onli:
Thank @GuillaumeValadas. In my test this works. In the pathological case with the - at the end the logic mode was disabled, and with proper search terms it worked, also other logic mode searches worked fine. I think this is definitely an improvement.

@garvinhicking Let's merge this! :)

garvinhicking:
@onli I trust your decission, go ahead :)

PR #853, opened 2024-08-16
serendipity_sendMail: terminate headers with CRLF.

### LINKED ISSUE #852: Comment notification mails unreadable due to missing encoding
### Problem description

s9y 2.5.0 on PHP 8.2.20

Comment notifications mails are shown unreadable in mail clients when encoded as Base-64. 

This is due to missing headers, especially "Content-Transfer-Encoding". Those headers are present in theory, but are appended as continuation lines (lines starting with whitespace) to another header.

Real-life example (slightly anonymised):

```
Subject: [Netz - Rettung - Recht] [Bewilligung notwendig] Neuer Kommentar [...]
From: "User" <account@mail.example>
 Reply-To: "User"<other.account@mail.example>
 X-Mailer: Serendipity/2.5.0
 X-Engine: PHP/8.2.20
 Message-ID: <081efa1ffe105d94af3352b006baf93e@netz-rettung-recht.de>
 MIME-Version: 1.0
 Precedence: bulk
 Content-Type: text/plain; charset=UTF-8
 Auto-Submitted: auto-generated
 Content-Transfer-Encoding: base64
Date: Fri, 16 Aug 2024 13:24:09 +0200
```

All headers from "Reply-To" up to "Content-Transfer-Encoding" have leading whitespace, so they are concatenaed to form this very long header:

```
From: "User" <account@mail.example> Reply-To: "User"<other.account@mail.example> X-Mailer: Serendipity/2.5.0 X-Engine: PHP/8.2.20 Message-ID: <081efa1ffe105d94af3352b006baf93e@netz-rettung-recht.de> MIME-Version: 1.0 Precedence: bulk
 Content-Type: text/plain; charset=UTF-8 Auto-Submitted: auto-generated Content-Transfer-Encoding: base64
```

I have reproduced that behaviour by commenting on old entries on my blog that are moderated.

### Probable cause

Those mails are sent via `serendipity_sendMail() `in `functions.php`.

All those headers with leading whitespace are added to the `$maildata['headers']` array, and then PHP's `mail()` function is called:

```
return mail($maildata['to'], $maildata['subject'], $maildata['message'], implode("\n", $maildata['headers']));
```

So all headers from the array - starting with "From" - are converted to a string, separeted with "\n" (LF). But mail headers should end with "\r\n" (CRLF). See RFC 5322:

```
2.2.  Header Fields

   Header fields are lines beginning with a field name, followed by a
   colon (":"), followed by a field body, and terminated by CRLF

```

So current versions of PHP - or Exim, my MTA - see "LF" as folding white space, but not as the end of a header line.

This is new to me. I had no problems with s9y 2.3.x on PHP 7.3.x. As the relevant code hasn't changed since 2.3.x, AFAIS, that's probably due to changed behaviour in PHP (or my mail server).

### Fix

Change the relevant line to

```
return mail($maildata['to'], $maildata['subject'], $maildata['message'], implode("\r\n", $maildata['headers']));
```

I can confirm that this change fixes the problem.

As the change makes s9y standard compliant, there should be no side effects.

th-h:
serendipity_sendMail() will concatenate additional headers to a string separated with LF (\n). That is not standard compliant, as RFC 5322 demands CRLF (\r\n).

This has real life implications at least on PHP 8.2.20 with Exim 4.96, as all additional headers are appended to one single long header line, losing encoding information and more.

Change LF to CRLF here.

Fixes #852.

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -20,6 +20,9 @@ Version 2.6-alpha1 ()
 
    * Block access to .inc php files via .htaccess
 
+   * Make serendipity_sendMail() standard compliant by using CRLF
+     instead of LF to separate headers.
+
 Version 2.5.0 (February 13th, 2024)
 ------------------------------------------------------------------------
 
--- a/include/functions.inc.php
+++ b/include/functions.inc.php
@@ -564,7 +564,7 @@ function serendipity_sendMail($to, $subject, $message, $fromMail, $headers = NUL
     }
 
     if (!isset($maildata['skip_native']) && !empty($maildata['to'])) {
-        return mail($maildata['to'], $maildata['subject'], $maildata['message'], implode("\n", $maildata['headers']));
+        return mail($maildata['to'], $maildata['subject'], $maildata['message'], implode("\r\n", $maildata['headers']));
     }
 }
 


garvinhicking:
Great, looks plausible.
Could you add an entry to docs/NEWS alongside this change?

th-h:
Yeah, sorry. I'm a bit rusty.

garvinhicking (approved):

PR #855, opened 2024-10-08
bugfix: pg_escape_string() in serendipity_db_escape_string() works again

HQJaTu:
Docs: https://www.php.net/manual/en/function.pg-escape-string.php

In PHP 8.1+ there is no default connection. A mandatory connection must be given as a parameter. This PR restores PostgreSQL support for S9y.

Not sure if PHP <8.1 must be supported. If yes, PHP_MAJOR_VERSION and PHP_MINOR_VERSION could be checked to determine which interface of `pg_escape_string()` needs to be used.

### DIFF
--- a/include/db/postgres.inc.php
+++ b/include/db/postgres.inc.php
@@ -94,6 +94,11 @@ function serendipity_db_escape_string($string) {
     if (is_null($string))
         return $string;
 
+    if (PHP_MAJOR_VERSION < 8 || (PHP_MAJOR_VERSION == 8 && PHP_MINOR_VERSION < 1))
+        # Last supported version is PHP 8.0
+        return pg_escape_string($string);
+
+    # From PHP 8.1 onwards
     return pg_escape_string($serendipity['dbConn'], $string);
 }
 
--- a/include/db/postgres.inc.php
+++ b/include/db/postgres.inc.php
@@ -89,7 +89,9 @@ function serendipity_db_reconnect() {
  * @return  string   output string
  */
 function serendipity_db_escape_string($string) {
-    return pg_escape_string($string);
+    global $serendipity;
+
+    return pg_escape_string($serendipity['dbConn'], $string);
 }
 
 /**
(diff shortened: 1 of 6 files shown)


onli:
Great :)

> Not sure if PHP <8.1 must be supported. If yes, PHP_MAJOR_VERSION and PHP_MINOR_VERSION could be checked to determine which interface of pg_escape_string() needs to be used.

It would be nice to support older PHP versions. It makes upgrades for old blogs a lot easier, and so far we were still compatible I think.

HQJaTu:
Checked after the commit. Correction: PHP 8.1 is still supported by security updates.
https://www.php.net/supported-versions.php

My own agenda is to move into PHP 8.2. When stable status reached, then bump into 8.3 to see what breaks.

include/db/postgres.inc.php
```diff
@@ -89,7 +89,17 @@ function serendipity_db_reconnect() {
...
 function serendipity_db_escape_string($string) {
-    return pg_escape_string($string);
+    global $serendipity;
+
+    if (is_null($string))
+        return $string;
+
+    if (PHP_MAJOR_VERSION < 8 || (PHP_MAJOR_VERSION == 8 && PHP_MINOR_VERSION < 1))
+        # Last supported version is PHP 8.0
+        return pg_escape_string($string);
```
onli, line 99: What do you think about this instead (untested)?

```
if (version_compare(PHP_VERSION, '8.1.0', '<')) {
    # Versions before 8.1 give the connection by default
    return pg_escape_string($string);
}
```

That's how the core would usually express that expression, and I think it's a bit clearer, isn't it?

onli:
Hi :) I did not see the progress here, sorry.  Thank you for changing it to `version_compare`, judging from the code that part looks good to me now!

But now there are some additional commits in this branch that probably ended up here by accident, like https://github.com/s9y/Serendipity/pull/855/commits/5c75b730fb836def4936c1fe56b2c0d76665f20e. Could you separate that, so we can merge this PR first? You could either rebase the branch to pick only the commits we need here or just open a new PR, whatever is easier.

The other commits could then in a PHP 8.2 fixes PR. Those I could actually test, and then we can pinpoint better if something broke.

onli:
I could not fix this PR, so instead the relevant commit was reused in https://github.com/s9y/Serendipity/pull/882. Will close here.

PR #856, opened 2024-10-13
Bugfix: Using PHP 8.2 dynamic properties removed

HQJaTu:
PHP 8.2 removes support of dynamic class properties. Compensate the change on selected plugins.

While looking at those changes a pattern emerges.
Maybe having `$markup_elements` and `$dependencies` in parent class should be a faster and cleaner fix. While at it drop usage of way-back-obsoleted `let`.

### DIFF
--- a/include/db/postgres.inc.php
+++ b/include/db/postgres.inc.php
@@ -89,7 +89,9 @@ function serendipity_db_reconnect() {
  * @return  string   output string
  */
 function serendipity_db_escape_string($string) {
-    return pg_escape_string($string);
+    global $serendipity;
+
+    return pg_escape_string($serendipity['dbConn'], $string);
 }
 
 /**
--- a/include/db/postgres.inc.php
+++ b/include/db/postgres.inc.php
@@ -91,6 +91,9 @@ function serendipity_db_reconnect() {
 function serendipity_db_escape_string($string) {
     global $serendipity;
 
+    if (is_null($string))
+        return $string;
+
     return pg_escape_string($serendipity['dbConn'], $string);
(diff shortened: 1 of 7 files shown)


HQJaTu:
Any progress on this?

plugins/serendipity_event_textile/serendipity_event_textile.php
```diff
@@ -251,7 +252,7 @@ function _process_tag_php($text)
                 '<font color="',
                 '</font>',
                 "\n ",
-                '� '
+                '� '
```
garvinhicking, line 255: unsure what this character would be (probably some ISO-8859-1 one?!) and what it would then maybe break. Could also be a UTF-8 BOM in ISO-8859-1?!

include/db/postgres.inc.php
```diff
@@ -89,7 +89,12 @@ function serendipity_db_reconnect() {
  * @return  string   output string
  */
 function serendipity_db_escape_string($string) {
-    return pg_escape_string($string);
+    global $serendipity;
+
+    if (is_null($string))
+        return $string;
+
+    return pg_escape_string($serendipity['dbConn'], $string);
```
garvinhicking, line 97: Wasn't this another issue/patch? Unsure how it affects existing setups...?

onli: Yes, there sadly seem to be some commits that got mixed together :/

onli:
I could not fix this PR, so I transported the relevant commit to https://github.com/s9y/Serendipity/pull/881. Will close here.

Issue #847, opened 2024-07-17
serendipity_event_textile: not compatible to php 8.1


bauigel:
I did an update to s9y latest git-master and now i get a lot of error-messages. 

    Notice: preg_split(): Passing null to parameter #3 ($limit) of type int is deprecated in /plugins/serendipity_event_textile/lib3/src/Netcarver/Textile/Parser.php on line 2181.
    
    Notice: preg_match_all(): Passing null to parameter #2 ($subject) of type string is deprecated in /plugins/serendipity_event_textile/serendipity_event_textile.php on line 164.
    
    Notice: str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /serendipity_event_textile/serendipity_event_textile.php on line 180.

The fix is easy for php 8.1 but I'm not sure if there is something to add to be downwards compatible. Maybe there is an update for textile to?

onli:
Yes, https://github.com/textile/php-textile should be the new upstream for the textile plugin, and that one had a very recent release.

bauigel:
With current master I get an warning

> str_replace(): Passing null to parameter #3 ($subject) of type array|string is deprecated in /var/www/virtual/bauigel/bernd.distler.ws/plugins/serendipity_event_textile/serendipity_event_textile.php on line 169

Fix would be to attach ?? '' to $eventData

`$eventData[$element] = str_replace('&quot;', '"', $eventData[$element] ?? '');`

onli:
I pushed a slightly different fix to master, same idea. Works for you as well?

onli:
Awesome, thanks. Then I'll close here.

PR #864, opened 2025-01-19
Update textile to v4, remove old parsers

onli:
This PR reworks the textile plugin to only include the current version of the parser (marked as compatible with PHP >= 5.3, but I did not test that) via composer. It also removes the old textile parser we included as lib1, lib2 and lib3, partly under the assumption that they won't work with current PHP versions anyway.

Note that the approach of bundling the textile parser in the plugin directory is debatable. Why do this for a core plugin, isn't that the job of bundled_libs/? On the other hand, should this plugin be a core plugin in the first place? Textile seems to have fallen out of favor

Related to https://github.com/s9y/Serendipity/issues/847, reporting incompatibility notices with PHP 8.1.


onli:
Merged to facilitate testing by @bauigel

PR #865, opened 2025-01-19
Update to smarty 5

onli:
Composer update to smarty 5 + necessary fixes to get the smarty classes working again.

Tested superficially in my test blog only so far. https://smarty-php.github.io/smarty/5.x/upgrading/#backwards-incompatible-changes-to-custom-plugins lists a lot of removals, I'd be surprised if this was everything to do.


garvinhicking:
Wow!!! You're on a roll. Smarty5 and the outlook on changes was what killed me.

If this is working this would be insanely great and a HUGE step towards keeping s9y alive.

I love you! 🥹

garvinhicking:
As for testing:

- we should clear templates_c to ensure no compuled templates from old smarty are around
- do custom registered smarty functions still work?
- check all our setendipity_smarty stubs if they are still callable/executable
- see if code modifying references (instead copied variables) is around and how to deal with that? I think there were some things were smarty modified variables and later PHP code must access those modified variables

But one step at a time :)

onli:
Love you too ;)

You caused me a bad conscience though, the testing was _very_ preliminary. Only the later commit fixed entry creation, after your comment.

> we should clear templates_c to ensure no compuled templates from old smarty are around

It gets repopulated properly :)

> do custom registered smarty functions still work?

Yes, the system works. In fact it's necessary to add the php functions to that system that were provided by **Serendipity_Smarty_Security_Policy** before. I did that in https://github.com/s9y/Serendipity/pull/865/commits/732f6711415a4be2790f83d3262e4260e4ee1cdd.

> check all our setendipity_smarty stubs if they are still callable/executable

This could maybe need some unittests... In general those seem to work though, otherwise the dev blog would not render.

> see if code modifying references (instead copied variables) is around and how to deal with that? I think there were some things were smarty modified variables and later PHP code must access those modified variables

Right, that was removed:

> You cannot use plugins that expect a parameter by reference anymore. PHP-function such as reset(), prev(), next() and end() can be registered as plugin, but they won't work because they expect a parameter by reference and Smarty will try to pass it by value. 

I did not notice usage of that, ofc I might have just missed it.

onli:
@garvinhicking I added sometest cases now. They don't cover all those functions, but they do test that a few custom modifiers work, as well as one PHP function and modifier that were broken after the smarty 5 upgrade initially. Let's merge this?

garvinhicking:
IMO this is the only way forward. We need to adopt here or get left behind too much. I trust in your work so far, couldn't check/test it out myself though. Generally this looks like a good approach, so.... fingers crossed? Leeeeroyyyyyyyyyyy

garvinhicking (approved):

onli:
Worst case we revert and try again :) I merged it

Issue #797, opened 2022-11-30
Quicksearch vanishes


onli:
The quicksearch sidebar plugin vanishes. Not in all themes, not in Skeleton. But it disappears in old blogs. My theory: 2k11 hides the plugin in favor of its own search, via the **sidebar.tpl**:

```
{if $is_raw_mode}
<div id="serendipity{$pluginside}SideBar">
{/if}
{foreach from=$plugindata item=item}
{if $item.class != "serendipity_plugin_quicksearch"}
    <section class="sidebar_plugin clearfix {cycle values="odd,even"} {$item.class}">
        {if $item.title != ""}
        <h3>{$item.title}</h3>
        {/if}
        <div class="sidebar_content clearfix">{$item.content}</div>
    </section>
{/if}
{/foreach}
{if $is_raw_mode}
</div>
{/if}
```

This now gets repeated. Solution: Add a sidebar.tpl without the quicksearch check to all themes?

garvinhicking:
Maybe a better solution could be to patch 2k11s siderbar tpl and add a specific check if $serendipity['theme'] == '2k11', and only then hide the search? If a fallback theme is used, it should keep its distinct name, I think?

garvinhicking:
(Adding a sidebar.tpl to themes means we can only patch it for available themes, and any custom-made theme would still be broken in that regard. And people would have to update external themes first...)

onli:
Alternative solution: Maybe we should remove this check? I imagine that this is really surprising behaviour for a lot of situations, and to limit it to 2k11 now would also be surprising. 

Does hiding the quicksearch really still make sense for a theme which templates are the foundation of almost all other themes? Is it necessary to hide the quicksearch sidebar plugin for some specific usecase (e.g. extending the built-in search with the livesearch plugin?) or is this just about not doubling the search - and if it is just for that, can't user just uninstall it if is unnecessary?

onli:
Fixed in https://github.com/s9y/Serendipity/pull/866 :)

PR #866, opened 2025-01-24
Limit integrated searchbox to 2k11 (#797)

onli:
Fixes old themes, by making it possible to use the quicksearch plugin and not having to add styles for the integrated searchbox. All includes new themes won't be affected since they have their own index.tpl.

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,6 +1,9 @@
 Version 2.6-alpha1 ()
 ------------------------------------------------------------------------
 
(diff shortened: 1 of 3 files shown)


onli:
@garvinhicking, this we discussed way back when I reported the bug in #797. I will just merge it in a bit if you can't find time for this (though it's very little code :) )

garvinhicking (approved):
Lgtm ;)

PR #870, opened 2025-02-10
Remove & reference declaration from db_query to avoid warning

### LINKED ISSUE #848: error from mysql.inc.php
With current git-master I get an error from mysql. 

    Notice: Only variable references should be returned by reference in /include/db/mysqli.inc.php on line 101.

onli:
Also adjusts the code that calls the function with a prepended &, as that would have caused a new warning. Fixes #848.

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,5 +1,7 @@
 Version 2.6-alpha1 ()
 ------------------------------------------------------------------------
+   * Avoid PHP 8 warning about incurrect reference return of db_query
+     function
 
    * Fix option to set a different theme as fallback option by setting
      Engine: $themename in a themes info.txt
--- a/include/db/mysqli.inc.php
+++ b/include/db/mysqli.inc.php
@@ -58,7 +58,7 @@ function serendipity_db_in_sql($col, &$search_ids, $type = ' OR ') {
  * @param   boolean     If true, the executed SQL error is known to fail, and should be disregarded (errors can be ignored on DUPLICATE INDEX queries and the likes)
  * @return  mixed       Returns the result of the SQL query, depending on the input parameters
  */
-function &serendipity_db_query($sql, $single = false, $result_type = "both", $reportErr = false, $assocKey = false, $assocVal = false, $expectError = false) {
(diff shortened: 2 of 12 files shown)


onli:
@garvinhicking I don't currently understand the idea of the & function declaration. What was the idea here, how could database query results be passed as reference and thus mutated? I assume I'm missing something, this is not my area of expertise.

onli:
Okay, I looked into this a bit more. My research so far:

https://www.php.net/manual/en/language.references.return.php shows clearly how this is supposed to work (now). A class variable is shared by returning it as a reference. This example does not apply here, as this returned with `$rows` a variable the function itself created, so the alias we are creating here does not point to something shared.

Why is the code like this? I do not know, but it is 19 years old, https://github.com/s9y/Serendipity/blob/80e14346a55f439896dd01553ed4e30db12d93d3/include/db/mysql.inc.php#L61 already returned the reference. My guess is that this was a performance optimization aimed at PHP 4 - it likely reduced memory usage back then by not having to copy the function return value to a new variable.

Will this change have a negative impact? According to my code search in additional_plugins, almost all calls to `serendipity_db_query` already use the regular assignment (`$somevar = serendipity_db_query(...)`). Exception is serendipity_event_userprofile, which uses at https://github.com/s9y/additional_plugins/blob/ab6addd0f8f708b09b651a365dc05641ac8c7134/serendipity_event_userprofiles/serendipity_event_userprofiles.php#L276, but looking at the code I see now purpose to that. And https://github.com/s9y/additional_plugins/blob/master/emerge_spartacus.php#L22 copies the function declaration, so this would have to be changed later when spartacus is build with a new s9y version. 
It is possible I missed a function call here though, and ofc plugins outside of spartacus might be unknowingly affected.

---

I intend to merge this. It silences a notice, should make the code easier and I do not expect bugs, as the unclear semantics of the reference passing probably made it meaningless in PHP > v5. I'll give it a few more days for potential feedback though :)

garvinhicking (approved):
Thanks for taking care.

Yes, this was a micro optimization to hint PHP at "knowing" it shall not re-create intermediate variables in memory for passing function results around, so no "copy by variable".

PHP8 is so much beyond needing such an optimization so for legibility, this change should be totally fine I believe.

PR #873, opened 2025-02-17
Remove jQuery migrate 

onli:
Removes jQuery migrate and replaces the parts of the backend JS that does not work without it. 

A followup to #687


onli:
Done in the backend as far as I can see. TODO: Test relevant plugins.

onli:
I'm through with the plugins. Almost all of them worked without issues with the new jQuery version or did not rely on it in the first place. 

I provided patches for the PHP issues and other breakage I noticed anyway, apart from  serendipity_event_cal - the maintainer left the project and the plugin is too work intensive to keep it aligned - and serendipity_event_podcast - similarly too big. Both of them did not seem specifically broken though, apart from warnings that can still be ignored for now.

Themes for a second time next, then a rework of the tab interface code in the backend, and then this should be done.

onli:
Fitvid does not seem to be a problem, as tested with 2k11 and an entry with a Youtube embed. And the tabs in the backend now use details/summary, for a less hacky solution. I think this is done! I'll leave this open for a bit to a check, otherwise merge for more testing by users of the master branch.

PR #875, opened 2025-02-23
Replace (int) parser conversion with spl_object_hash

onli:
The (int) conversion does not work anymore since PHP 8 changed the xml parser from being a resource to being an object. Hardcoding it to a number, the initial fix, destroys the logic of having multiple parsers available and was not done everywhere. The object hash approach should be a valid replacement.

Draft, as still to be tested.

### DIFF
--- a/bundled-libs/XML/RPC.php
+++ b/bundled-libs/XML/RPC.php
@@ -245,7 +245,7 @@ function XML_RPC_se($parser_resource, $name, $attrs)
 {
     global $XML_RPC_xh, $XML_RPC_valid_parents;
 
-    $parser = (int) $parser_resource;
+    $parser = spl_object_hash($parser_resource);
 
     // if invalid xmlrpc already detected, skip all processing
     if ($XML_RPC_xh[$parser]['isf'] >= 2) {
@@ -374,7 +374,7 @@ function XML_RPC_ee($parser_resource, $name)
 {
     global $XML_RPC_xh;
 
-    $parser = (int) $parser_resource;
+    $parser = spl_object_hash($parser_resource);
 
     if ($XML_RPC_xh[$parser]['isf'] >= 2) {
         return;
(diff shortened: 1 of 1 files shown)


onli:
I added tests and made them pass. 

Main question now: This approach of patching our old XML/RPC.php, is that the way we want to continue? The lib was already patched for s9y, so I went that route. But also because I saw no clear maintained replacement - https://pear.php.net/package/XML_RPC points to the abandoned https://pear.php.net/package/XML_RPC2. 

If there is a candidate, that could be suggested instead. https://gggeek.github.io/phpxmlrpc/ looks good, but would be a different API - probably easy enough for serendipity_event_weblogping, but I did not look into serendipity_event_xmlrpc, where the server component could change a lot.

bundled-libs/XML/RPC.php
```diff
@@ -1433,7 +1433,9 @@ function parseResponse($data = '')
 
         $encoding = $this->getEncoding($data);
         $parser_resource = xml_parser_create($encoding);
-        $parser = 1;
```
garvinhicking, line 1436: This was a scalar value, maybe copy pasta mistake?

bundled-libs/XML/RPC.php
```diff
@@ -1530,7 +1532,7 @@ function parseResponse($data = '')
                 $r = new XML_RPC_Response($v);
             }
         }
-        $r->hdrs = preg_split("@\r?\n@", $XML_RPC_xh[$parser]['ha']);
+        $r->hdrs = preg_split("@\r?\n@", array_key_exists('ha', $XML_RPC_xh[$parser]) ? $XML_RPC_xh[$parser]['ha'] : '');
```
garvinhicking, line 1535: I would prefer:
```suggestion
        $r->hdrs = preg_split("@\r?\n@", $XML_RPC_xh[$parser]['ha'] ?? '');
```
for legibility

bundled-libs/XML/RPC.php
onli: That was a mistake I did in https://github.com/s9y/Serendipity/commit/a6619f77cd840f2c1888524052aa3c07c0b645b4#r152820624. I needed a replacement for `(int) $parser_resource;`, and though this could just be hardcoded to a number. That might even be feasible, but not when changing it only at one of the places that creates `$parser` like this. The replacement is the point of this PR, to use `$parser = is_resource($parser_resource) ? ((int) $parser_resource) : spl_object_hash($parser_resource);` everywhere.

PR #876, opened 2025-02-26
Enable custom parameters for .webp and .avif thumbnails

onli:
The thumbnail code assumed that images are .gif, .png or .jpg, and we defaulted to quality settings for .jpg thumbnails. But for both .avif and .webp, other quality settings are preferable (and possible other additional parameters).

This code changes the code for $serendipity['imagemagick_thumb_parameters'] so that it can be an array, with the image mime type as key and their specific parameters as values. It also sets sane defaults for the gd path for both formats, which also enables thumbnails in both formats when imagemagick is not used.


onli:
My uberspace dev server does not support .avif in the gd path yet, so .avif thumbnail creation with gd is untested. Should work fine though. 

I also got a gateway timeout right now when uploading images after switching to imagemagick, but the thumbnails got properly created. An uberspace hickup? I will check again later.

Issue #637, opened 2019-09-17
Denial of Service via template cache and disk space


hannob:
It is possible to fill up the disk space available to an s9y installation. The only precondition is that URL rewriting is enabled.

s9y will generate a cache entry for every URL *including 404 URLs* in templates_c. So by simply calling many invalid paths on a host with an s9y installation one can fill up the disk space. The speed of this attack depends on how much blog content there is, as the 404 page contains the blog front page.

With a simple curl-based attack (code below) I can fill up around 3 megabyte per second. This attack is not optimized, by parallelizing and maybe using http2 one can probably do this much faster. But this optimized version already allows filling up the disk with ~ 10 gigabyte per hour.

There should be some mechanism to avoid caching 404 pages (or alternatively they should all access the same cache entry).

Attack script:
```
#!/bin/bash

target="$1"

urls=""
for i in $(seq 1 1000); do
	for j in $(seq 1 1000); do
		urls="$urls $target/$i-$j"
	done
	curl -sk $urls > /dev/null
done
```

(I herebly allow redistributing this script licensed as CC0 in case anyone cares)

hannob:
I'm wondering if the cache is actually needed at all.

I have done some simple benchmarks (using recursive "wget -l 1 -r --domain [host]"), with existing cache, no existing cache (cleaning templates_c dir before) and with disabled cache. I don't see any significant difference at all. Always takes something between 9 and 10 seconds, if there is any speed advantage by the cache then it's hardly measurable.

This is certainly not the most sophisticated benchmark. But it makes me wonder if the cache is based on outdated assumptions (after all PHP got a lot of performance in recent years) and maybe it's just the best to get rid of it?

onli:
Hi @hannob. I'm just realizing I'm not sure which cache you mean. There actually are at least two: the caching layer we added, simple_cache, and the php files Smarty is producing. Thinking about it you probably mean the simple_cache?

[...]

Issue #345, opened 2015-05-29
Speedup s9y by using a cache


onli:
See the [forum-thread](http://board.s9y.org/viewtopic.php?f=11&p=10443063&sid=610c75731adf30ecb844d04f5fa0d8fb#p10443063).

onli:
Please test the cache by setting [$use_internal_cache](https://github.com/s9y/Serendipity/blob/0125f1539e4305d6ee0de0368672a49f40f9b3ac/serendipity_config.inc.php#L454) to true.

garvinhicking:
Nice, I'm really excited about feedback for this. Could you also document this in docs/NEWS for testers?

onli:
Sure, done

onli:
@garvinhicking Based on positive first feedback, I'd like to add a real config option and then close here. Ok?

garvinhicking:
Fine by me!

PR #883, opened 2025-04-02
Use custom cache implementation

onli:
Replaces voku/simple-cache with some custom PHP code that caches either to APCu or to templates_c/.

The goal here is to achieve a simplification. Cache libraries would provide the ability to use systems like memcached or redis as a cache, but we never fully activated that for voku/simple-cache nor do we have a good way to provide a configuration for that, if those servers can't be autodetected. But the libs can produce issues: voku/simple-cache specifically hasn't seen a release in over 2 years and produces warnings under PHP 8.4, other prominent cache libraries do not support PHP 7.x (which might make upgrades more difficult for some users).

S9y already used an internal API for the cache, so I could just replace the wirings to the external caching library with code that does the caching directly. This also made it possible to finally address https://github.com/s9y/Serendipity/issues/637 at the caching layer, by adding a check for the cache size and a cleanup if there are too many cache items.


onli:
What do you think? Does this approach make sense?

hannob:
I would like to repeat what I already said in multiple comments in #637 

I have disabled the cache on my blogs. I tried measuring performance differences with and without cache. I didn't see any. My impression is that, due to the performance improvements in PHP itself, the cache provides little to no advantage on modern systems.

My recommendation would be to first make sure the cache is actually providing a performance advantage, and then consider how to improve it. And if it's not providing any advantage, it should be removed.

onli:
@hannob I think you are right. I tested the cache from this PR now on a very weak server (a PogoPlug, ARMv5 single core with 256MB ram). Summary of my tests (what I think is the core metric, always two results):

`ab -n 1000 -c 5 -C "somecookie=rawr" http://192.168.1.57/`:

No internal cache: [195ms, 195ms] TPR 
Internal cache (filesystem): [187ms, 187ms] TPR
Internal cache (APCu): [188ms, 188ms] TPR

`ab -kc 10 -t 30 http://192.168.1.57/` 

No internal cache: [202ms, 211ms] TPR 
Internal cache (filesystem):  [198ms, 210ms] TPR
Internal cache (APCu): [200ms, 209ms] TPR

`wrk -t5 -c10 -d30s http://192.168.1.57/`

No internal cache: [150, 150] requests
Internal cache (filesystem): [100, 160] requests
Internal cache (APCu): [154, 154] requests

`ab -n 200 -c 20 -C "somecookie=rawr" http://192.168.1.57/`

No internal cache: [200ms, 201ms] TPR 
Internal cache (filesystem):  [192ms, 206ms] TPR
Internal cache (APCu): [194ms, 195ms] TPR

`ab -n 200 -c 50 -C "somecookie=rawr" http://192.168.1.57/`

No internal cache: [224ms, 224ms] TPR 
Internal cache (filesystem):  [216ms, 208ms] TPR
Internal cache (APCu): [222ms, 227ms] TPR

---

I'd summarize this as there being an advantage for the cache, but it is small. The other metrics like latency etc also do not change that impression much. Sure, might be different with a bigger database and more plugins, but on the other hand, this is already on a very weak server where I expected the cache to matter the most. It's also surprising that the APCu cache is not clearly better than the file system cache - after all, the file system is on an USB stick! Note that the Zend opcache had a huge effect on performance here.

I also wasn't able to reproduce the result from back then, when the cache did not help performance under low load, but massively prevented timeouts on high server loads. Couldn't find such parameters with these two tools on that setup.

---

In a first experiment, moving the cache out of `serendipity_fetchEntries` to `serendipity_printEntries` did improve performance on that server a lot. Though that will cut a lot of the dynamic functionality, that might be worth it and give the cache purpose. I'll experiment with this a bit more and report in a second comment.

onli:
So, the current state is worth considering:

`ab -n 200 -c 20 -C "somecookie=rawr" http://192.168.1.57/`

No internal cache: [203ms, 203ms] TPR
Internal cache (filesystem): [103ms, 110ms] TPR
Internal cache (APCu): [99ms, 109ms] TPR

`ab -n 200 -c 30 -C "somecookie=rawr" http://192.168.1.57/`

No internal cache: [194ms, 195ms] TPR
Internal cache (filesystem): [99ms, 149ms] TPR
Internal cache (APCu): [98ms, 109ms] TPR

`wrk -t5 -c10 -d30s --latency http://192.168.1.57/`

No internal cache: [159, 159] requests
Internal cache (filesystem): [189, 305] requests
Internal cache (APCu): [299, 307] requests

I'll add full logs as gists later. 

Price of the performance are the plugin hooks, as to be expected. Worthwhile to note, also to have this written down for later: This is pretty much how the cache was intended in the first place. [The triggering forum post](https://board.s9y.org/viewtopic.php?f=11&t=20345) talked about a static text cache and I proposed the printEntries layer, also in https://github.com/s9y/Serendipity/issues/345. That was part of what landed with https://github.com/s9y/Serendipity/commit/0125f1539e4305d6ee0de0368672a49f40f9b3ac#diff-97fe7ae78f8f65b7e8b44e56fe7c44c9b17e05a7f193c9f9a7b4c70c07fd2fd0, but I later removed the printEntries cache part in https://github.com/s9y/Serendipity/commit/bf213b752df884d80767f9b9e9649d330731f9d8. Sadly with no reference to a discussion :/ It is possible that the printEntries cache as it existed back then was ineffective, as it still called plugin hooks - so not enough measuring.

@hannob Could you give the current approach a try with your benchmark?

PR #887, opened 2025-04-09
Remove remnants of /archives as filesystem directory

onli:
As s9y works currently, the **/archives**directory is not needed. `/archives` is used as a path in the URL, but the entry is completely created dynamically from the database. There was old code to store entries as files under **/archives/**, but that code was either never or never in the last 20 years in use. But the installer still checked that **/archives** exists and is writeable. 

This PR removes the check and the few other references. It will not removes the directory on existing installations, after all user might have stored files there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,6 @@ plugins/additional_plugins*
 /.htaccess
 *.db
 private/*
-archives/*
 *~
 DEADJOE
 /.settings
--- a/include/admin/installer.inc.php
+++ b/include/admin/installer.inc.php
@@ -268,18 +268,6 @@ if ( (int)($serendipity['GET']['step'] ?? null) == 0 ) {
         }
     }
 
-    if ( is_writable($basedir . 'archives/') ) {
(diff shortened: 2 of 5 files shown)


garvinhicking:
Phew. I hardly remember. I think there were some plugins (additional_plugins) that wrote files there for "persistence". Maybe one of the caching ones. So something not like "templates_c" which is considered temporary and be removable at all times, but also not "user generated content" like "uploads".

Maybe the external plugins would need to be searched for it. Personally I would keep it around because I think removing it can do more bad than good?

But I'm also fine with trying it out and seeing what breaks. Users can still create the directory (and existing installations will not remove it, so.....)

onli:
ah, I didn't think of that. Looks like serendipity_event_downloadmanager uses archives/ as its exchange directory. serendipity_event_backendrss stores some file there, and so did blogpdf (but that one could be ignores, sitting under /alpha). 

`serendipity_event_backendrss` could probably just be moved to templates_c/, but downloadmanager looks like that might break things there - assuming the plugin works currently.

onli:
Okay, I think it's not worth it. Thanks @garvinhicking :)

PR #893, opened 2025-04-29
Make entry_id an unsigned integer

fasterit:
Fixes: PHP Fatal error:  Uncaught mysqli_sql_exception:
 Incorrect integer value: '' for column `serendipity`.`serendipity_spamblocklog`.`entry_id`

Issue with php 8.4

### DIFF
--- a/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php
+++ b/plugins/serendipity_event_spamblock/serendipity_event_spamblock.php
@@ -1718,7 +1718,7 @@ class serendipity_event_spamblock extends serendipity_event
             default:
                 $q = sprintf("INSERT INTO {$serendipity['dbPrefix']}spamblocklog
                                           (timestamp, type, reason, entry_id, author, email, url,  useragent, ip,   referer, body)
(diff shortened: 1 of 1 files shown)


onli:
Is it possible the issue is somewhere else?

I don't see how changing `%s` to `%u` can help here. The  `$id` should always be a positive number, and the string representation of an int is the same in both cases. E.g.

```
echo sprintf('%s, %u', 1000, 1000);  # => 1000, 1000
```

So I assume if this fails, `$id` is set to a negative number here. Which skimming the code is surpring, `$id` seems to always come from `$eventData['id']`, which should be positive as well.

Maybe you already debugged where the ill-fitting number is coming from? That could be a bug of its own.

fasterit:
No, I just got the error message when running against php 8.4 (Trixie) and MariaDB/InnoDB. Changing the %s -> %u fixed it. I did not investigate further.
/DLange

onli:
Hm, okay. Let's merge this then. But please keep an eye out for the origin bug, there should be no negative id there to trigger the mysql exception you got.

fasterit:
will do && thanks

PR #903, opened 2025-06-24
Proposal: Set default value of captchas to off

onli:
As a minimally invasive change, I would like to change the default of the captchas to disable them. 

Captchas are no deterrent to bots anymore, or only block the lowest effort bot out there. But they still are an accessibility problem. To reduce their harmfulness, we can disable them by default, which might make some blogs out there not enable them.

As an alternative proposal, maybe a follow-up, I would replace them with a question. Like "What is 1 + 1", maybe even a small selection of questions . This will also not block all bots, but it would have the same effect of blocking the lowest effort bots. Probably would be a drop-in replacement.

Alternative to the alternative proposal: Just remove them completely. We do have bayes and spamblock bee and the other spam measures of the spamblock plugin, options that work better and are less harmful.

What do you think? @garvinhicking ?

### DIFF
--- a/plugins/serendipity_event_spamblock/ChangeLog
+++ b/plugins/serendipity_event_spamblock/ChangeLog
@@ -1,3 +1,5 @@
+1.89.7: Change the default for captchas to off, for accessibility
+
 1.89.6: More PHP 8 compatibility fixes
 
 1.89.3: PHP 8 compatibility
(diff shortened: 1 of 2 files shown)


garvinhicking:
I don't like captchas either but I believe letting them solve text riddles is easier than running OCR. S9y is no commin target so hardly spambots will be optimized to recognize our "custom" ones. Using AI is too cost intensive for these bots.

Probably a honeypot is more effective, catching comments that are submitted too quickly.

Just disabling captchas will probably let in a LOT of spam. Then again, those users can re-enable captchas.

Tl;dr: feel free to do this :)

Issue #907, opened 2025-07-16
Simeplepie needs update for PHP 8.4


onli:
The version of the bundled-lib simplepie is outdated and breaks on PHP 8.4. It is used in the core plugin serendipity_plugin_remoterss if the feed type is set to Atom. Additionally, the spartacus plugins serendipity_plugin_socialbookmarks and serendipity_event_aggregator seem to provide their own outdated copies.

The project seems to be active, https://github.com/simplepie/simplepie/, so we can probably just update the bundled lib (via composer). And then make the plugins use the bundled lib. Though there might be some code changes needed.

Bonus goal: Simplepie supports RSS and Atom, so there seems to be no need to use both in the remoterss plugin. Merging both paths should get rid of some code duplication. On the other hand, Onyx works fine so far. Might not be worth it if we don't want to drop that dependency completely now.

PR #909, opened 2025-07-22
Update Simplepie to 1.8.1 and move it to composer

onli:
Simplepie was bundled in the very old version 1.2.1 and broke under PHP 8.4 (and possible earlier). Simeplepie is used in the core only in the remoterss plugin, and then only if an Atom feed (instead of an RSS feed or an opml file) is read. The remoterss plugin was updated to work with the new version. There is also a redirect in the old `simplepie/simplepie.inc` location, to minimize possible breakage - though like remoterss showed, the new version is not completely backwards compatible anyway. Composer does not download the single file version of simplepie used before, thus the redirect.

Because simplepie can read Atom and RSS this also removes the code path in the remoterss plugin that used Onyx for RSS, instead also using simplepie here. That worked in my test just fine. Some warnings about the content of html node elements, but those wouldn't be shown in a production mode.

Fixes https://github.com/s9y/Serendipity/issues/907.


onli:
Okay, test passes and I also tested this in my dev setup. Note that the https://github.com/s9y/Serendipity/pull/909/commits/cec9842fd7223081d554c4536a9bac1000abf0e7 fix for the testcase under PHP 7.4 is noteworthy, I did not expect this. Given that this should be part of the next beta release I'll go ahead and merge.

PR #928, opened 2025-10-17
Fix social plugin etc not having access to cached entries

onli:
The social plugin accesses entries in the frontend_header hook by accessing `$eventData['smarty']->tpl_vars['entries']`. But with the reworked cache, that smarty variable was not set when the entry came from the cache. This restores that access method by setting the variable on cache hit.


onli:
Strangely enough, while this worked for having the og-elements reappear with image and link, title and description are missing in my blog for my latest entry. It works in the others - I'll try to reproduce first.

Issue #656, opened 2019-10-18
Brute force protection for serendipity login


stephanbrunker:
I don't have that much experience, but for a project of mine I have realized an protection against brute force attacks on the login. Because we all know that too much users use passwords which doesn't have that much complexity, one solution is to limit the number of login attempts in a given time frame. For example, after 20 attempts block the user for one hour. This is the same as in Girocard protection which has only 2^13 possibilities, but is blocked after three wrong attempts. 

To realize that, you'll need two additional fields in the user data: last login and login attempts, everything else is not that complicated.

But I really don't know if that is necessary or not.

fe-hicking:
I‘d be all for it! Sounds good to have. However, how do I protect myself that someone who tries to login with my username and uses a bot that tries ever 5 minutes, so a constant lock would happen?

onli:
You are worried about someone locking a user out? The block should be combined with the IP, not just the username

fe-hicking:
Yeah exactly. But if its locked to an IP it can be brute forced again with botnets easily, which have many IPs to outsource to.

stephanbrunker:
I never thought of the lock-out scenario. But in that case if someone is laying siege to your site, you have really a problem. It depends somewhat on the users involved: If the password is strong enough which should be the case for administrators, it shouldn't be necessary. But for the page I built I didn't hash but encrypted the passwords in the database that I can decrypt them and check (I know, that is dubious too, but the users are my employees and the data is critical). The reality is unfortunately that common users mostly use passwords which aren't very secure. I just looked into Joomla and there is only a timeout for the "forgot password" function and the login attempts are logged, but not restricted. On the other hand, the Internet connection by Deutsche Telekom blocks you after only three or five attempts (easily if you put the wrong numbers in your router and it tries again) and locks you out until midnight. So, really hard to say if this feature would be an improvement in site security or a risk for DDOS attacks.

onli:
Having the integrated intelligence to throttle/ban an attacker would be nice and a net win in my opinion. The ddos scenario is way harder to implement (from the attacker perspective) than a simple bot trying to guess the password. So if we can make the simple attack scenario harder we gain a bit of security in the more realistic scenario.

Note: fail2ban (+password authentication on apache level) would already allow that now.

th-h:
+1, as long as it's configurable ... or tied to the attacker's IP address.

[...]

PR #941, opened 2025-11-11
Send and validate a second factor to finish login

onli:
Ask the user to enter a code sent by email to his user email address, in a second step after the regular login.

Draft status, still needs:

- [x] to be made configurable.
- [x] Might need additional logic to persist the 2fa status, that the code was entered and validated. 
- [x] Email template and language constants.
- [x] Logic check, that the way this is added to the flow is effective enough to block users that entered a valid password but not the second factor

Follow-up to https://github.com/s9y/Serendipity/issues/656#issuecomment-543939203

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,6 +1,10 @@
 Version 2.6-beta2 ()
 ------------------------------------------------------------------------
 
+   * Provide option for a second factor on login. Currently this
+     consists of a code sent to the autor's email that has to be entered
+     on login.
+
    * 2k11 now shows examples for its multiple date format options in the
      theme's configuration
      


stevleibelt:
If the HMAC is stored as a one-to-many relation to the user account, the setup on the server should be possible by a dedicated 2fa user settings page or section.
Either you can select to have a E-Mail or your own 2fa application.
Via E-Mail the server is calculating the token and sends it to the user.
Via 2fa-app, the token is created on a local device.
Anyway, after the regular login with the username and the password, the second factor is asked and the validation logic is the same.

From my point of view, creating the user configuration page, hook into the login-process and finding a well tested third party library could be the harder areas.

PR #960, opened 2026-04-06
Let users provide a list of valid hosts for autodetected base URLs 

onli:
Give an option to secure the http host detection when using the autodetect base URL mode.

Thanks to [@mabjr33 ](https://github.com/mabjr33) for the suggestion.


mattsches (commented):
I added some comments.

mattsches (approved):

PR #961, opened 2026-04-06
Send 2FA token iff user does not already have an incoming token

onli:
Without this check the user got a new 2fa code on every login page load.

### DIFF
--- a/include/functions_config.inc.php
+++ b/include/functions_config.inc.php
@@ -499,13 +499,20 @@ function serendipity_setAuthorToken() {
     $_SESSION['author_token'] = $hash;
 }
 
+/**
+ * Send a login code to the user's email adress, if the user has not currently already an active
+ * login code.
+ */
 function serendipity_send2faCode() {
     global $serendipity;
-    $secondFactor = bin2hex(random_bytes(3));
-    serendipity_cacheItem($serendipity['serendipityUser'] . '_2faCode', $secondFactor, 60 * 15);
-    $subject = sprintf(SECOND_FACTOR_MAIL_TITLE, $serendipity['serendipityUser']);
-    $message = sprintf(SECOND_FACTOR_MAIL, $serendipity['serendipityUser'], $secondFactor);
-    return serendipity_sendMail($serendipity['serendipityEmail'], $subject, $message, $serendipity['blogMail']);
+    $storedSecondFactor = serendipity_getCacheItem($serendipity['serendipityUser'] . '_2faCode');
+    if (! $storedSecondFactor) {
+        $secondFactor = bin2hex(random_bytes(3));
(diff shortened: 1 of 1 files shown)


include/functions_config.inc.php
```diff
@@ -499,13 +499,20 @@ function serendipity_setAuthorToken() {
...
     global $serendipity;
-    $secondFactor = bin2hex(random_bytes(3));
-    serendipity_cacheItem($serendipity['serendipityUser'] . '_2faCode', $secondFactor, 60 * 15);
-    $subject = sprintf(SECOND_FACTOR_MAIL_TITLE, $serendipity['serendipityUser']);
-    $message = sprintf(SECOND_FACTOR_MAIL, $serendipity['serendipityUser'], $secondFactor);
-    return serendipity_sendMail($serendipity['serendipityEmail'], $subject, $message, $serendipity['blogMail']);
+    $storedSecondFactor = serendipity_getCacheItem($serendipity['serendipityUser'] . '_2faCode');
+    if (! $storedSecondFactor) {
+        $secondFactor = bin2hex(random_bytes(3));
+        serendipity_cacheItem($serendipity['serendipityUser'] . '_2faCode', $secondFactor, 60 * 15);
```
mattsches: Can we store the code only after successfully sending the mail? I think `serendipity_sendMail` returns mail success, so if mailing the code fails, the 2faCode should not be stored.

Also, writing to cache could also fail, and we shouldn't send the mail then, but this is a edge case, I guess.

mattsches (commented):
I suggested some changes.

include/functions_config.inc.php
onli: Yes, we can do that :) I changed it around.

Writing to cache can fail, right, but it really should be very rare.

mattsches (approved):

include/functions_config.inc.php
```diff
@@ -508,11 +508,16 @@ function serendipity_send2faCode() {
...
         $subject = sprintf(SECOND_FACTOR_MAIL_TITLE, $serendipity['serendipityUser']);
         $message = sprintf(SECOND_FACTOR_MAIL, $serendipity['serendipityUser'], $secondFactor);
-        return serendipity_sendMail($serendipity['serendipityEmail'], $subject, $message, $serendipity['blogMail']);
+        if (serendipity_sendMail($serendipity['serendipityEmail'], $subject, $message, $serendipity['blogMail'])) {
+            return serendipity_cacheItem($serendipity['serendipityUser'] . '_2faCode', $secondFactor, 60 * 15);
+        } else {
+            return false;
+        }
     }
+    return false;
```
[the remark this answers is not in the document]
onli: Both ways of thinking work I'd say: One could pick true here since at that point there was already a valid 2FA token stored. I picked false since at that point no new token was generated or sent out.

PR #968, opened 2026-04-16
Update comments.inc.php for compatibility with PHP 8.3

qbi:
When you want to access all comments, S9Y runs into an HTTP 500: `PHP Fatal error:  Uncaught DivisionByZeroError: Division by zero` in line 260.

The comparison of strings and integers has changed in PHP 8. See https://www.php.net/manual/en/migration80.incompatible.php

Previously the comparison in line 260 was `0 == 'all'` -> true. Whith the changes the comparison now is `0 == 'all'` -> false. Which leads to the above mentioned error message.

I changed the code in a way that `$commentsPerPage` has either an interger or`COMMENTS_FILTER_ALL` as value and the comparison now checks type and value. So in the case of all comments the comparison is `'all' === 'all` -> true.

This should fix the code.

### DIFF
--- a/include/admin/comments.inc.php
+++ b/include/admin/comments.inc.php
@@ -6,7 +6,11 @@ if (IN_serendipity !== true) {
 
 $data = array();
 
-$commentsPerPage = (int)(!empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10);
+$commentsPerPage = !empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10;
+if ($commentsPerPage != COMMENTS_FILTER_ALL) {
+    $commentsPerPage = (int)($commentsPerPage);
+}
+
 $summaryLength = 200;
 
 $errormsg = array();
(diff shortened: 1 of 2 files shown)


include/admin/comments.inc.php
```diff
@@ -6,7 +6,11 @@
 
 $data = array();
 
-$commentsPerPage = (int)(!empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10);
+$commentsPerPage = !empty($serendipity['GET']['filter']['perpage']) ? $serendipity['GET']['filter']['perpage'] : 10;
+if ($commentsPerPage != COMMENTS_FILTER_ALL) {
```
mattsches, line 10: Thank you, good catch and good fix!

My suggestion would be to change the first section as follows.
This introduces a stricter (type-safe) check against `COMMENTS_FILTER_ALL`.

```php
if ($commentsPerPage !== COMMENTS_FILTER_ALL) {
    $commentsPerPage = (int)$commentsPerPage;
}
```

What do you think?

mattsches (changes requested):

onli:
Surprisingly complicated code, originally. I assume it was never intended to rely on `0 == 'all'` being true, but rather it was missed that the comparison was strange.

Alternative fix would be to always cast COMMENTS_FILTER_ALL when used in the comparison, so also on line 260 and 274, like 

```
$pages = ($commentsPerPage == (int)COMMENTS_FILTER_ALL ? 1 : ceil($totalComments/(int)$commentsPerPage));
```

Then this becomes a `0 == 0` when `$serendipity['GET']['filter']['perpage']` was set to `COMMENTS_FILTER_ALL` (which will be a string like 'all', and casts to 0)

Not sure that's better, but wanted to understand the bug and have this shared :)

onli:
I prepared the NEWS entry for this change. @qbi, please let us know if this is the way you want to be credited, or under your full name? If you could add @mattsches suggestion this would be great, otherwise I'll merge this as is the next days. In any case, thanks for finding and fixing this bug.

PR #969, opened 2026-04-26
Fix 404 on cached page loads from plugins (i.e. staticpages)

fasterit:
After upgrading to Serendipity v2.6.0, I had the following issue:
When requesting a static page that is linked to e.g. /pages/contact.html it works (HTTP status 200). Requesting the same page again (i.e., reloading the page in the browser, with the s9y_ session cookie set) is results in a HTTP 404 error and the text "The requested page could not be found (404). This is the default page.". The default page below then is the static page that is gets correctly loaded from the database and rendered. Just the 404 and the resulting error message are ... wrong.

Fixed this by massaging the caching logic and re-firing the events.
Not sure whether there is a better place to fix the logic.
@onli ... please take a look.

/DLange

### DIFF
--- a/docs/NEWS
+++ b/docs/NEWS
@@ -1,6 +1,9 @@
 Version 2.7-alpha1 ()
 ------------------------------------------------------------------------
 
+   * Fix error message and 404 status code on plugin pages when cache
+     was active (thanks to @fasterit/DLange)
+
 Version 2.6.0 (April 10th, 2026)
 ------------------------------------------------------------------------
 
--- a/include/functions_entries.inc.php
+++ b/include/functions_entries.inc.php
@@ -1088,6 +1088,24 @@ function serendipity_printEntries($entries, $extended = 0, $preview = false, $sm
 
         $cached = serendipity_getCacheItem($cache_key);
         if ($cached && $cached !== false) {
+            if ($use_hooks && ($serendipity['view'] ?? '') === '404') {
+                // When the current view is '404', a plugin (e.g. staticpage) may have populated this
+                // cache entry by handling the URL as a clean page.  On cache hits the entry_display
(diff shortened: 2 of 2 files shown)


onli:
Thanks for noticing this. I can confirm this bug in my blog. 

I wonder whether an alternative fix would be to not cache those plugin pages. Or to cache them in a way that these side effects can be triggered without re-running the event_hook - running it too often is what killed performance of the prior cache implementation. Maybe this concept of a plugin page can be something that is more structured, more than the clean_page hook result, and then we can detect it easier/not have the side effects be triggered like this in the first place.

fasterit:
I think not caching them is a bad option as they may be frequently requested and creating them is quite expensive.

The other part of your considerations seem reasonable. I won't dig into this but if you have the time, please do.

We can merge this one and replace it with something better later, too.

onli:
You are right, let's work with what we have instead of waiting for a more complicated solution. But I really want to avoid the second event hook call. So I propose this instead: https://github.com/s9y/Serendipity/compare/master...prevent-staticpage-404 - based on your fix, detect plugin pages with a second cache key entry and then apply the side effects.

(I wanted to push onto this PR and messed up the commands, so now it's a separate branch but still with your commit at the start, so I hope that's okay.)

Could you give this a look and ideally a test run? And also have a look at the NEWS entry, whether this is your preferred attribution.

fasterit:
Your change with the `||pluginpage` cache entries works fine. Tested :heavy_check_mark: .

@fasterit is fine for credits or DLange, we have both in `docs/NEWS` already.
DLange was (and is) just not available on github, so used the company name instead.

fasterit:
I have cleaned up your commits and added them squashed to the PR branch ^.
You can use that for merging, if you want to.

onli:
Thanks a lot!

PR #971, opened 2026-04-29
Bypass the cache when comments are previewed or submitted otherwise

fasterit:
After 680d70aa comment previews are ... getting the cache page back and comment submission feedback is chewed up, too.
Make these requests bypass the cache.


onli:
Thanks again

PR #972, opened 2026-04-30
Bypass cache for entry/staticpage preview, don't poison cache when we can easily avoid it by checking bypass_cache

fasterit:


onli:
Could you add a line to the docs/NEWS file describing the change? Then it can be merged together more easily.

fasterit:
This is still fixing fallout from

  * Fix error message and 404 status code on plugin pages when cache
     was active (thanks to @fasterit/DLange)

I don't think there's any user value in detailing where it broke things along the way.

If you think differently, please propose a line and I'll happily add it.

onli:
No, that's alright then. I missed the connection. Thank you for continuing with the fixes.

PR #975, opened 2026-07-11
support SQL execution after aquiring a new database connection

mmitch:
I have switched my s9y instance from MySql/MariaDB to SQLite in 2024 and while doing the update to s9y 2.6.0 I've just stumbled about a local patch with the notice "upstream this". Here it is:

I've looked a bit into tuning the SQLite database access simply by measuring page load times. With some simple changes I got a measurable speedups. Database tuning in SQLite is done by sending some `PRAGMA` SQL commands after the connection to the database was is aquired.  
Because s9y has no way to do this that I know of, I've added a new global configuration variable `dbSqlOnNewConnection` that makes it possible to execute some SQL statements automatically on every new database connection.  The variable is optional (so new changes neccessary on existing installations) and accepts either a single SQL command as a String or multiple commands as an Array of Strings.  
(Somehow I was not able to combine multiple `PRAGMA`s separated by `;` into a single SQL statement to be executed, so I added the array form as an alternative.)

This is the configuration I currently use (excerpt from my `serendipity_config_local.inc.php`)  – while this might not be fit for copy and paste because it is tuned to my database size and setup, it shows how this is supposed to work:

```php
    // set DB connection parameters after connect
    $serendipity['dbSqlOnNewConnection'] = [
        'PRAGMA journal_mode = WAL;',
        'PRAGMA synchronous  = FULL;',          // eigentlich ON, aber zu wenig Schreibzugriffe, dann lieber richtig
        'PRAGMA journal_size_limit = 2097152;', // 2MB, die ganze DB ist ja keine 16 MB groß...
        'PRAGMA mmap_size    = 16777216;',      // 16MB; bisher war die ganze DB zwischen 9 und 13 MB groß, sollte reichen
        'PRAGMA cache_size   = 512;',           // 512x4k -> 2MB, das ist ziemlich genau der Default von -2000 = 2000k :)
        'PRAGMA busy_timeout = 10000;',         // max. 10 Sekunden auf Connection warten (intern mit exponentiellem Backoff)
        # 'PRAGMA foreign_keys = ON;'            // FIXME: es gibt aktuell keine Fremdbeziehungen, liegt das am DB-Import oder macht s9y das immer ohne?
    ];
    /*
      defaults on my system:
      journal_mode = delete
      synchronous = 2
      journal_size_limit = -1
      mmap_size = 0
      cache_size = -2000
      busy_timeout = 60000
      foreign_keys = 0
    */

```

For the available `PRAGMA`s see: https://sqlite.org/pragma.html

I think this should at least be useful on every installation with the SQLite database backend to be able to tune the database access, but it might also come in handy with other databases or for some completely different purposes.

(I've also added some lines of code to the admin interface to show the current SQLite `PRAGMA` settings, but it's simply thrown into the category page and not properly formatted. I'll omit this for now.)

If you find this now configuration variable useful, please accept this pull request.
It's rebased against the current master branch and should apply without problems.

### DIFF
--- a/include/db/pdo-sqlite.inc.php
+++ b/include/db/pdo-sqlite.inc.php
@@ -104,6 +104,13 @@ function serendipity_db_connect() {
                                  'sqlite:' . (defined('S9Y_DATA_PATH') ? S9Y_DATA_PATH : $serendipity['serendipityPath']) . $serendipity['dbName'] . '.db'
                              );
 
+    // Handle additional connection setup if set
+    if (isset($serendipity['sqlitePragmas'])) {
+        foreach($serendipity['sqlitePragmas'] as $sql) {
+            serendipity_db_query($sql);
+        }
+    }
+
     return $serendipity['dbConn'];
 }
 
--- a/include/db/sqlite.inc.php
+++ b/include/db/sqlite.inc.php
@@ -57,6 +57,13 @@ function serendipity_db_connect()
         $serendipity['dbConn'] = $function($serendipity['dbName'] . '.db');
     }
 
+    // Handle additional connection setup if set
+    if (isset($serendipity['sqlitePragmas'])) {
+        foreach($serendipity['sqlitePragmas'] as $sql) {
+            serendipity_db_query($sql);
+        }
+    }
+
     return $serendipity['dbConn'];
 }
 
--- a/include/db/sqlite3.inc.php
+++ b/include/db/sqlite3.inc.php
@@ -53,6 +53,13 @@ function serendipity_db_connect()
     // SQLite3 doesn't support persistent connections
     $serendipity['dbConn'] = sqlite3_open((defined('S9Y_DATA_PATH') ? S9Y_DATA_PATH : $serendipity['serendipityPath']) . $serendipity['dbName'] . '.db');
 
+    // Handle additional connection setup if set
+    if (isset($serendipity['sqlitePragmas'])) {
+        foreach($serendipity['sqlitePragmas'] as $sql) {
+            serendipity_db_query($sql);
+        }
+    }
+    
     return $serendipity['dbConn'];
 }
 
--- a/include/db/sqlite3oo.inc.php
+++ b/include/db/sqlite3oo.inc.php
@@ -53,6 +53,13 @@ function serendipity_db_connect()
     // SQLite3 doesn't support persistent connections
     $serendipity['dbConn'] = new SQLite3((defined('S9Y_DATA_PATH') ? S9Y_DATA_PATH : $serendipity['serendipityPath']) . $serendipity['dbName'] . '.db');
 
+    // Handle additional connection setup if set
+    if (isset($serendipity['sqlitePragmas'])) {
+        foreach($serendipity['sqlitePragmas'] as $sql) {
+            serendipity_db_query($sql);
+        }
+    }
+
     return $serendipity['dbConn'];
 }
 


onli:
The tests failing is unrelated and a helpful info, that's something to fix for PHP 8.1 caused by https://github.com/s9y/Serendipity/commit/9c593c341e0c7e9993f21e94eb1384667207ba0b.

For the PR, is that really useful for other databases than SQLite? To my knowledge those are configured serverside and won't need such a hook. I wonder whether we want to make this a mechanism specific so SQLite, run some default pragmas at https://github.com/s9y/Serendipity/blob/9e21ad371f89937fe78151881eb955928c41822d/include/db/pdo-sqlite.inc.php#L100 with a hook like this to change them/add additional Pragma commands. Call it `$serendipity['sqlitePragmas']`?

mmitch:
I don't know if it is useful for other databases. My idea was to keep it generic for future use-cases I can't think of yet.

But I have looked around a bit: PRAGMAs only seem to exist on SQlite, MariaDB has nothing like that.
The only _other_ SQL I could think of running at connection aquisition would be incrementing a counter counting the connection aquisitions – but that is quite far-fetched.

I'll update it like you proposed:
* move it to the SQLite backend
* change the name to `sqlitePragmas`

With the name change to plural I'd also drop the string/array distinction and switch it to array only. No need to overcomplicate things.

onli:
If you update the branch to the current master (you can pull it, sitcch back to this branch and then do `git merge master`) the tests should pass now.

mmitch:
I did a rebase, but the test actions don't seem to run again automatically.
@onli could you please trigger a manual run?

onli:
Might be because of the force-push, there is only one commit in the branch, and that's still the old one?

mmitch:
It's a new one (commit c1184d1 instead of 83a5c0f), but of course mostly the same: The parent-commit has changed and I made a small change to the commit message.

Looking at the triggers in the workflow definition, it's the same as in my own GitHub workflows. I regularly force-push rebased commits and other in-place changes on feature branches in my projects and the workflows then run as expected.
I can't see what's wrong here.

Because I'm no project menber I can neither look deeper into it nor start a run manually.
If all else fails I could try to open another pull-request :-/

onli:
I'll send you an invite. 

But my comment was only meant to help, the pipeline doesn't have to pass now, after all this functionality is still to be changed.

mmitch:
Ah, there was a run 14 hours ago, but I only saw it in my inbox:
https://github.com/s9y/Serendipity/actions/runs/29167974167
There was one error:
https://github.com/s9y/Serendipity/actions/runs/29167974167/job/89821360759

D'oh, I forgot to remove the array/string distinction as promised.
I'll get to it!

onli:
> There was one error:
> https://github.com/s9y/Serendipity/actions/runs/29167974167/job/89821360759

I triggered that run, but don't understand why the error persists, I thought I fixed it with https://github.com/s9y/Serendipity/commit/343cb695e5d55d6d7e32d473aeea3e2d750fe748

mmitch:
I've updated the PR again, now
* the variable is called `sqlitePragmas` like you have proposed
* it is only enabled for SQLite databases
* it handles arrays only, no extra "single string" variant

I have found that there are multiple database drivers for SQLite:
* `include/db/pdo-sqlite.inc.php`
* `include/db/sqlite.inc.php`
* `include/db/sqlite3.inc.php`
* `include/db/sqlite3oo.inc.php`

I have updated all of them, but my blog runs with `pdo-sqlite` so I have only tested that one.

onli:
> I have updated all of them, but my blog runs with pdo-sqlite so I have only tested that one.

That's fine, I think that ought to be distilled down to that one anyway.

PR #977, opened 2026-07-23
Do not override login form user with HTTP auth user

### LINKED ISSUE #967: Non-admin user cannot login after update to 2.6.0 on PHP 8.4
I run a German Serendipity on PHP 8.4 for many years now. Two days ago I updated from 2.5.0 to 2.6.0.

I have one Administrator user and an Editor user (Redakteur). With the admin user I have no issues - repeated logout/login just works without resetting password.

The issue is with the editor user: When I first tried to login after the update I got the "failed" message. So I changed the password using the admin account, entered the old and new password and saved the user. I can confirm that this change worked by using MySQL client to take a look into the users table - the password hash has been changed.

After setting the new password, I could login with the editor ONCE. After logging out and re-login, I got the "failed" message again! I cleared cookies, and also tried login with another browser, but cannot login.

Then I changed the password a second time using administrator, but that did not work. Editor cannot login anymore, neither with an old nor changed password.
I even tried to copy the admin password hash to the editor user with MySQL, so that both have the same password hash. But this also didn't work. My edtor user cannot login.

Finally I installed the "forgot password" plugin, but this seems to not work correctly, as I don't see a "forget password" link anywhere. So I am really stuck with my editor user now.

fantasya-pbem:
fixes #967

### DIFF
--- a/serendipity_config.inc.php
+++ b/serendipity_config.inc.php
@@ -132,7 +132,7 @@ if (!isset($serendipity['expose_s9y'])) {
 // If set to true (in serendipity_config_local.inc.php) this prevents using imap_8bit
 // functions to send a mail, and use base64 encoding instead
 if (!isset($serendipity['forceBase64'])) {
-    $serendipity['forceBase64'] = false; 
+    $serendipity['forceBase64'] = false;
 }
 
 // Should IFRAMEs be used for previewing entries and sending trackbacks?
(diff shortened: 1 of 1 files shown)


fantasya-pbem:
My proposal is to only use HTTP auth user/pass if no POST variables have been submitted from logiin form, therefore this additional "if" around the existing code.

Removed some superfluous spaces, too.

onli:
That looks completely reasonable. It fixes the bug in your installation?

fantasya-pbem:
Turns out : no. I think I'll have a deeper look into this file and debug it locally.

PR #982, opened 2026-09-03
update release script

### LINKED ISSUE #978: remove .github directory from installation ZIP
The released ZIP file contains a "hidden" `.github/` directory to be deployed in the s9y installation.

I have manually removed the directory and then the upgrade wizard warned about these two missing files:

* `.github/dependabot.yml` beschädigt oder verändert: Überprüfung fehlgeschlagen
* `.github/workflows/php_tests.yml` beschädigt oder verändert: Überprüfung fehlgeschlagen

This happens with both versions 2.6.0 and 2.6.1.

I think the the directory and both files should be removed from the installation ZIP because they serve no purpose in a deployed blog (not even as documentation like `README`, `LICENSE` or similar files do).

mmitch:
Please see issue #978 for the background.

This PR tweaks various things about the `bundled-libs/create_release.sh` script:
- Remove .github/ directory from release archive
- Don't include create_release.sh in the release archive
- Remove special CVS handling from create_release.sh
- Fix ShellCheck findings in create_release.sh
- Fix directory check in create_release.sh
- Abort create_release.sh when a .git directory is detected

Please see the individual commits for further information.

### DIFF
--- a/bundled-libs/create_release.sh
+++ b/bundled-libs/create_release.sh
@@ -89,8 +89,12 @@ else
             echo "   - $2/bundled-libs/create_release.sh [766]"
             chmod 766 "$2/bundled-libs/create_release.sh"
             
+            echo "   - $2/.github [remove]"
+            rm -rf "$2/.github"
+
             echo "   - $2/tests [remove]"
             rm -rf "$2/tests"
+
             echo "    [DONE]"
             echo ""
 
--- a/bundled-libs/create_release.sh
+++ b/bundled-libs/create_release.sh
@@ -86,8 +86,8 @@ else
             echo "   - $2/upgrade.sh [744]"
             chmod 744 "$2/upgrade.sh"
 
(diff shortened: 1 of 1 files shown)


bundled-libs/create_release.sh
```diff
@@ -41,18 +41,18 @@ else
         echo "         to bundle a new release version!"
         echo ""
         echo "Hit [ENTER] to continue, or abort this script (CTRL-C)"
-        read -n 1
+        read -r _
         gensums=0
-        which php > /dev/null
-        if [ $? -ne 0 ]
+        
+        if which php > /dev/null;
```
mmitch: This is condition is the wrong way around, it must be negated.
The error message should be displayed when `which` fails (no `php` binary found), currently the error message is shown when `php` is available.

I'll fix this by rebasing to keep the history clean.

mmitch: Fixed, new commit is 278692f16bebe60f80b7e415635c367052090c0e

